PR #190, opened 2013-06-24
Change streamMode to read-only on url_start

joostfaassen:
Here's a minor fix that allows opening read-only files through Gaufrette/StreamWrapper.

Without this fix, Gaufrette opens all streams as read/write before calling stat() on them.
Afaik this is not needed, only 'r' is sufficient. Without this fix, calling 'stat' on read-only files fails.


wysow:
@joostfaassen is this still relevant?

joostfaassen:
@wysow no, current master solves this problem. thanks!

PR #195, opened 2013-07-01
Don't override the region by default

stanlemon:
By default the client will come into the adapter with a region specified, and if the region is not override it will default to US-East anyhow.  By default the 'region' setting it means that set_region() will always be called on the client.  This can cause problems if allow_hostname_override() but more importantly if I set a different region on the client when I initially set it up it will get overridden unless I explicitly set the 'region' option.

Consider this example...

```php
<?php
$amazonS3 = new AmazonS3(array(
	'certificate_authority' => true,
	'key' => 'mygovcloudkey',
	'secret' => 'mygovcloudsecret',
));
$amazonS3->set_region(AmazonS3::REGION_US_GOV1);

// Despite setting this I MUST set 'region' below

$adapter = new Gaufrette\Adapter\AmazonS3($amazonS3, 'mybucket', array(
	'region' => AmazonS3::REGION_US_GOV1,
));
```

This change would mean that I could just do:
```php
<?php
$adapter = new Gaufrette\Adapter\AmazonS3($amazonS3, 'mybucket');
```

But I could also still supply 'region' if needed and override it accordingly.

This is a bigger deal if the S3 client is setup in some other scope and a developer is injecting it into Gaufrette and just assuming it's going to work.

Hope this makes sense, let me know if it needs further clarification.

### DIFF
--- a/spec/Gaufrette/Adapter/AmazonS3Spec.php
+++ b/spec/Gaufrette/Adapter/AmazonS3Spec.php
@@ -51,7 +51,6 @@ class AmazonS3Spec extends ObjectBehavior
             ->shouldBeCalled()
             ->willReturn(new \CFResponse('header', 'some content', 200))
         ;
-        $service->set_region(Argument::any())->shouldBeCalled();
 
         $this->setMetadata('filename', $options);
         $this->read('filename')->shouldReturn('some content');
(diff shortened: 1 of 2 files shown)


wysow:
Thx @stanlemon and sorry for the delay ;)

stanlemon:
No worries, thanks for merging!

Issue #250, opened 2014-01-08
Local adapter not implementing ListKeysAware


garak:
I see that Ftp adapter implements ListKeysAware just listing files/dirs and doing some regexp.
Can't Local adapter do just the same?

PR #252, opened 2014-01-09
add localKeys to Local adapter

garak:
Implement ListKeysAware for Local adapter. See also issue #250

---
This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/252)

### DIFF
--- a/spec/Gaufrette/Adapter/LocalSpec.php
+++ b/spec/Gaufrette/Adapter/LocalSpec.php
@@ -57,6 +57,12 @@ class LocalSpec extends ObjectBehavior
         $this->keys()->shouldReturn($expectedKeys);
     }
 
+    function it_lists_keys()
+    {
+        $expectedKeys = array('keys' => array('dir/file', 'filename'), 'dirs' => array('dir'));
+        $this->listKeys()->shouldReturn($expectedKeys);
+    }
+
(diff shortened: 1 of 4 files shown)


wysow:
@garak if this is still relevant, can you rebase this PR?

JoniJnm:
It's easy to use implements `ListKeysAware` in Local adapter only changing in the current `keys()` function `$this->directory` by `$this->computePath($key)`.
The problem is the function `listKeys` in [FileSystem](https://github.com/KnpLabs/Gaufrette/blob/master/src/Gaufrette/Filesystem.php#L159). I don't know why returns an array of "keys" and "directories".

It should return an array of keys. We can create maybe another function with "listDirs" or something. And the function should accept a `$recursive` function.

stof:
We cannot change the return type of an existing method of the filesystem class. This is a BC break.

JoniJnm:
That's way the milestone is 1.0

NiR-:
Indeed we can't do it for 0.x. But `keys` and `listKeys` are badly design from my point of view, so eventually they'll change (for the v1).

garak:
Being this PR too old, relevant issue closed, and hard-to-solve conflicting, I'm closing.
Thanks anyway for support

PR #289, opened 2014-05-14
Adding OpenStack adapter

wysow:
Adding an adapter for OpenStack, used for example for Hubic, OVH personal cloud storage.


wysow:
This will be added in the new Gaufrette organization in the near future, closing this here

PR #294, opened 2014-06-06
check if source is written/renamed correctly

### LINKED ISSUE #293: [cache] source rename/write fail
There is a serious bug in cache adapter -  if source has failed to write or to rename a file (eg. S3 invalid key in my case) following happens:
 - native aws exceptions is suppressed with catch return false
 - return false by `source` silently ignored
 - `cache` has been written
 - Filesystem reports write/rename was successful 
 - FAIL, file does not exist

[Gaufrette\Adapter\Cache::rename()](https://github.com/KnpLabs/Gaufrette/blob/master/src/Gaufrette/Adapter/Cache.php#L98)

[Gaufrette\Adapter\Cache::write()](https://github.com/KnpLabs/Gaufrette/blob/master/src/Gaufrette/Adapter/Cache.php#L108)

jaymecd:
Resolves #293 issue

### DIFF
--- a/spec/Gaufrette/Adapter/CacheSpec.php
+++ b/spec/Gaufrette/Adapter/CacheSpec.php
@@ -87,7 +87,7 @@ class CacheSpec extends ObjectBehavior
      */
     function it_rename_file_in_source_and_cache($source, $cache)
     {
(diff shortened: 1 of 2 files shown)


wysow:
Thx @jaymecd!

PR #303, opened 2014-08-12
Cleans up Local adapter docblocks and removes unnecessary use statement

woodsae:
Cleans up Local adapter docblocks and removes unnecessary use statement


AntoineLelaisant:
This PR is open for 3 years, I close it.

PR #307, opened 2014-09-24
Fixed AmazonS3 keys method

alex-pex:
The method didn't care about "directory" option.

I also modified phpspecs to reflect real "get_object_list" behaviour : an item in a subdir isn't listed alone, the subdir is listed before.

Travis build is OK

---
This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/307)

### DIFF
--- a/spec/Gaufrette/Adapter/AmazonS3Spec.php
+++ b/spec/Gaufrette/Adapter/AmazonS3Spec.php
@@ -461,7 +461,7 @@ class AmazonS3Spec extends ObjectBehavior
         $service
             ->get_object_list('bucketName')
             ->shouldBeCalled()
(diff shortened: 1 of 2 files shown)


NiR-:
Thanks for your contribution. Unfortunately we'll deprecate AmazonS3 soon (see #464), so I close it.

alex-pex:
@NiR Any alternative? We are using Gaufrette as an abstraction on S3 / Local storage

NiR-:
@alex-pex Yep, you can rely on [`AwsS3` adapter](https://knplabs.github.io/Gaufrette/adapters/aws-s3.html) rather than `AmazonS3`. It's the newer, maintained adapter for S3. And it should provide the same functionality as `AmazonS3`. Tell us if encounter any friction while migrating to it.

Issue #224, opened 2013-10-29
Add Descriptive Exceptions


jeremylivingston:
The Filesystem class currently throws a few \RuntimeExceptions during certain failures. I think these should be modified to throw more library-specific exceptions that extend RuntimeException. There are many different errors that could trigger a RuntimeException, so it would be nice if we could catch connection issues specifically.

I can add this if everyone is on board. I just wanted to get the group's feedback.
Thanks!

PR #308, opened 2014-09-25
Added possibility to throw specific exceptions at the adapter level

drgomesp:
This aims to solve #224, by adding the possibility of throwing a specific `\RuntimeException` from the adapter level. This can be a simple wrapped exception, like implemented on the AwsS3 adapter:

```php
try {
    $this->service->putObject($options);
    return strlen($content);
} catch (\Exception $e) {
    throw new \RuntimeException('Could not put object on S3.', 0, $e);
}
```

Or a full subtype exception of the `\RuntimeException`.

If this PR makes sense to you, I'll change all of the other adapters.

### DIFF
--- a/spec/Gaufrette/FilesystemSpec.php
+++ b/spec/Gaufrette/FilesystemSpec.php
@@ -181,10 +181,10 @@ class FilesystemSpec extends ObjectBehavior
     function it_fails_when_write_is_not_successful($adapter)
     {
         $adapter->exists('filename')->willReturn(false);
-        $adapter->write('filename', 'some content to write')->shouldBeCalled()->willReturn(false);
+        $adapter->write('filename', 'some content to write')->shouldBeCalled()->willThrow(new \RuntimeException());
 
         $this
-            ->shouldThrow(new \RuntimeException('Could not write the "filename" key content.'))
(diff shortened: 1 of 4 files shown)


wysow:
@drgomesp Sorry for the delay on this one... If I'm not mistaken every adapter should be updated to follow what you did on the S3 one right?

drgomesp:
@wysow no problem, and sorry for my delay.

Not necessarily, as the title suggests (add *possibility*). However, if better, we can do that right away on this PR.

stof:
This change is totally incomplete: adding the possibility of throwing exceptions on failure rather than returning false is changing the API of the adapter interface (which is a BC break btw). So any place consuming the adapter need to take into account the new way to handle errors too.

NiR-:
@drgomesp First of all thank you for this PR, and sorry for the very long delay. We will try to push forward the effort in order to release a stable version in the next weeks. Because of the heavy use of this package, the 0.x branch is currently a trap for us: even if it's not a stable one we should act like it is.

As stated by stof, throwing an exception in the S3 adapter would imply a modification of the current Adapter API and introduction of a BC break. We simply can't do this for now, and we need to postpone this to a later release.
Also we need to implement library-specific exceptions to fix #224.

PR #309, opened 2014-10-06
Added redis support

mlimaloureiro:


wysow:
@mlimaloureiro is this still relevant? If yes can you update the incomplete tests?

NiR-:
Since we have `php-cache`, I think this PR is not relevant anymore. Rather, it would be cool to have some integrations with cache adapters (need to investigate what need to be done here).
@akovalyov @docteurklein what do you think ?

PR #315, opened 2014-11-27
Fix rename in AwsS3 adapter

lapause:
- The source file was left on S3. Added deletion after the copy operation.
- Added metadata handling for destination key. By default, S3 sets the destination ACL to private during copy operation, so this allows user to easily make the renamed file public.

### DIFF
--- a/src/Gaufrette/Adapter/AwsS3.php
+++ b/src/Gaufrette/Adapter/AwsS3.php
@@ -111,7 +111,7 @@ class AwsS3 implements Adapter,
 
         try {
             $this->service->copyObject($options);
-            return true;
+            return $this->delete($sourceKey);
         } catch (\Exception $e) {
             return false;
         }
--- a/src/Gaufrette/Adapter/AwsS3.php
+++ b/src/Gaufrette/Adapter/AwsS3.php
@@ -110,7 +110,7 @@ class AwsS3 implements Adapter,
         );
 
         try {
-            $this->service->copyObject($options);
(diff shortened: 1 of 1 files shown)


NiR-:
@lapause Please can you update `AwsS3Test` ?

lapause:
I don't use PHPUnit, so I have no idea how to correctly mock the AWS client and produce a useful test :-/ Can you handle it?

src/Gaufrette/Adapter/AwsS3.php
```diff
@@ -110,8 +110,8 @@ public function rename($sourceKey, $targetKey)
         );
 
         try {
-            $this->service->copyObject($options);
-            return true;
+            $this->service->copyObject(array_merge($options, $this->getMetadata($targetKey)));
```
NiR-: I don't think it's necessary to call `getMetadata` here because [`getOptions`](https://github.com/KnpLabs/Gaufrette/blob/master/src/Gaufrette/Adapter/AwsS3.php#L275) already call `getMetadata`.

lapause: It's been a long time since is issued this patch but if I remember correctly without this call the renamed file was always private no matter what visibility option I used.

trq:
Is this going to get merged any time soon?

akovalyov:
@NiR- just tested this patch on real S3 bucket. I confirm that bug exists and is fixed by this PR.
LGTM for me.

PR #320, opened 2015-01-09
Allow Local adapter mkdir mode to cascade to its Stream

richardfullmer:
I'm setting the mode for ensuring directories get created properly in the `Local` adapter as `0777`, but it's not respected by the Stream implementation of the Local adapter.  This causes my apache user and the cli user not to be able to read/write the same set of files when using streams.

**Note:**  This is a small BC break because the mode that comes from the `Local` adapter is defaulted to `0777` and the mkdir command hardcoded in the `Stream\Local` is `0755`.  Let me know if this is a problem or it's acceptable as is.

---
This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/320)

### DIFF
--- a/src/Gaufrette/Adapter/Local.php
+++ b/src/Gaufrette/Adapter/Local.php
@@ -149,7 +149,7 @@ class Local implements Adapter,
      */
     public function createStream($key)
     {
-        return new Stream\Local($this->computePath($key));
+        return new Stream\Local($this->computePath($key), $this->mode);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


NiR-:
@richardfullmer Thank you for your contribution. Could you add a test case checking the chmod of the directory please?

NiR-:
I close this PR since I opened a new one, with requested tests. Thank you @richardfullmer for your contribution, credits will go to you :)

PR #322, opened 2015-01-22
Fixed an undefined index.

mgdigital:
The FTP raw directory listing begins with a "total" line, which is causing an undefined index in the FTP adapter.

### DIFF
--- a/src/Gaufrette/Adapter/Ftp.php
+++ b/src/Gaufrette/Adapter/Ftp.php
@@ -436,7 +436,7 @@ class Ftp implements Adapter,
                         'name'  => $infos[8]
                     );
                 }
-            } else {
+            } elseif (count($infos) >= 4) {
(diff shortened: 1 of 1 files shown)


mgdigital:
Fixes issue #323

wysow:
@mgdigital Can you provide a full example and a test for this? Thx.

mgdigital:
I'm afraid I don't have time to write a test, but you can see from my patch that the FTP adapter is not checking an array index exists before accessing it. 

I assume this bug may only happen in certain environments (this happened to me under Ubuntu 14.4). As described above the FTP directory listing begins with a "total" line which the FTP adapter is trying to read as if it were part of the main file listing.

ianfp:
I am also experiencing this bug. Can you merge this PR?

NiR-:
@mgdigital thanks :)

PR #324, opened 2015-02-04
AWS_S3: add better write content handling

vvh-empora:
to prevent big copy by value operations

---
This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/324)

### DIFF
--- a/src/Gaufrette/Adapter/AwsS3.php
+++ b/src/Gaufrette/Adapter/AwsS3.php
@@ -120,10 +120,27 @@ class AwsS3 implements Adapter,
     /**
      * {@inheritDoc}
      */
-    public function write($key, $content)
-    {
+    public function write($key, $content) {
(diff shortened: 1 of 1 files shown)


NiR-:
First of all, thank you @vvh-empora for your contribution, and sorry for the really big delay.

I'm reluctant to approve your patch because it would break adapters compatibility: resource would be properly handled only by AwsS3 adapter. In the other hand I agree that copying the whole file content in memory multiple times to upload a file suck when it just does not work.

I would rather consider two architectural changes:
* Add dedicated method for every type of content (filepath, resource, plain string).
* Leverage Stream concept with different type of streams, so every adapter would be responsible for supporting each type either in a dedicated or a generic way.

In any case, I added a note to our Github project so we keep in mind issues related to big files.

PR #330, opened 2015-03-26
Limit crawler files scope

visual74:
In case in Local.php-keys() have to get list on very big project, it's very long ! The better way is specify the scope before crawl.
My edits give to Filesystem this possibility.

### DIFF
--- a/src/Gaufrette/Adapter/Local.php
+++ b/src/Gaufrette/Adapter/Local.php
@@ -88,10 +88,15 @@ class Local implements Adapter,
     /**
      * {@inheritDoc}
      */
-    public function keys()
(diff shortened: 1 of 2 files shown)


stof:
This is wrong as you don't respect the interface anymore (and changing the interface is a BC break)

visual74:
Right ! Have you got a solution ?
```php
interface Adapter{
     /** Returns an array of all keys (files and directories)
     *
     * @return array
     */
    public function keys();
```
In my case "all keys" means more than 100K folders and files... Too hard for the server.

stof:
@visual74 implement ListKeysAware if you have an optimized way to apply the prefix

PR #334, opened 2015-04-12
Added stream feature to AwsS3 adapter. 

Yuav:
This remove issues with out of memory on large files, and optimize delivery path to clients.

### DIFF
--- a/src/Gaufrette/Stream/Local.php
+++ b/src/Gaufrette/Stream/Local.php
@@ -12,9 +12,9 @@ use Gaufrette\StreamMode;
  */
 class Local implements Stream
 {
-    private $path;
-    private $mode;
-    private $fileHandle;
+    protected $path;
+    protected $mode;
+    protected $fileHandle;
 
     /**
      * Constructor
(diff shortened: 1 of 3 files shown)


src/Gaufrette/Stream/Local.php
```diff
@@ -12,9 +12,9 @@
  */
 class Local implements Stream
 {
-    private $path;
-    private $mode;
-    private $fileHandle;
+    protected $path;
+    protected $mode;
+    protected $fileHandle;
```
stof: this looks wrong to me

Yuav: Why?

stof: Making things protected means they become extension points which need to be taken into account for backward compatibility. And extension points by inheritance are the worse ones regarding BC layers (we implement BC layers in Symfony since years, and the ones involving inheritance and BC for protected things have always been the worse ones to implement)

Yuav: So what's the fix? Copy everything over?

PR #335, opened 2015-04-15
make rename with AWS S3 adapter delete source file

shieldo:
A rename operation should remove the source file.

This already happens in the AWS S3 adapter for Flysystem: https://github.com/thephpleague/flysystem-aws-s3-v2/blob/master/src/AwsS3Adapter.php#L254

### DIFF
--- a/src/Gaufrette/Adapter/AwsS3.php
+++ b/src/Gaufrette/Adapter/AwsS3.php
@@ -111,10 +111,11 @@ class AwsS3 implements Adapter,
 
         try {
             $this->service->copyObject($options);
-            return true;
         } catch (\Exception $e) {
             return false;
         }
+
+        return $this->delete($sourceKey);
     }
 
     /**


src/Gaufrette/Adapter/AwsS3.php
```diff
@@ -111,10 +111,12 @@ public function rename($sourceKey, $targetKey)
 
         try {
             $this->service->copyObject($options);
-            return true;
         } catch (\Exception $e) {
             return false;
         }
+        $this->delete($sourceKey);
```
staabm: what todo when `delete` fails?

shieldo: Does Gaufrette have a clear idea about what it *should* do when delete fails - should this operation be atomic? (If not, then it probably should.)

The local adapter uses PHP's `rename` function which "attempts" to rename - unclear whether that operation is atomic. The v1 Amazon S3 adapter returns `false` if the copy operation succeeds but the delete fails (i.e. not atomic). Maybe this one should do the same as v1, but I think Gaufrette should be opinionated about what happens.

shieldo:
@staabm I've reworked the commit to have the same logical behaviour as the v1 S3 adapter.

src/Gaufrette/Adapter/AwsS3.php
staabm: regarding php `rename`:

> However, copy() is, unlike rename(), NOT useable for "atomic updates". Another process may actually access the destination file while copy() is working. In such a case, the other process with perceive the file as empty or with incomplete content ("half written").

see http://php.net/manual/de/function.rename.php

shieldo: @staabm I guess that refers to the content of the file, not the atomicity of the copy-then-delete operation, right?

staabm: @shieldo not sure. it reads like that with rename either the operation fails and only the initial source will stay as is or the new destination will be created while the source will be gone, which sounds like a atomic rename.

A copy+rename in 2 operations cannot be atomic using this PHP-APIs IIRC.

shieldo: I guess the question is what we do here, because obviously the behaviour should be as similar as possible.  I've changed the commit here to be logically the same as the v1 adapter - hopefully that's OK/ no worse than before?

staabm: I have no idea what the author of the lib thinks, so lets hear for their opinion.

the initial version and also the new changed form both make sense for me.

shieldo: @staabm np, thanks for the comments!

shieldo:
The PR #315 also makes this change but wraps in a change to do with metadata. So if that's merged, close this one.

NiR-:
Thank you @shieldo. Because of the metadata thing, I would prefer merging #315, I close this one.

PR #347, opened 2015-07-15
Fix infinite loop bug check directory with Slash and backslash

VanVan:
Check directory is valid with Slash and backslash, not only with slash.
php function dirname() used in this class can return backslash

### DIFF
--- a/src/Gaufrette/Adapter/Ftp.php
+++ b/src/Gaufrette/Adapter/Ftp.php
@@ -338,7 +338,7 @@ class Ftp implements Adapter,
      */
     private function isDir($directory)
     {
-        if ('/' === $directory) {
+        if ('/' === $directory || '\\' === $directory) {
             return true;
         }
 


VanVan:
This is not a windows specific problem. The problem still remain with Linux FTP server.

akovalyov:
@VanVan hey.
I am really sorry that I came to you so late. 
Anyway, let's try to get this PR done. Could you assist me in that?
I recently tried to reproduce this issue at https://github.com/akovalyov/gaufrette-integration-tests/blob/master/features/issues/%23347.feature and didn't succeed in it.  Could you add a little bit more of details - in which env does the issue appear?

NiR-:
@akovalyov Duplicated of #271. The problem seems to appears only on windows.

VanVan:
As you can see if you look at my pull request, I just add support of "\" separator used on Windows. This commit that I aleady added to my code used in production resolved all these problems.
I don't remember how to reproduce this bug, but the change is very short and can't add another bug, so you can merge it without hesitation :)

VanVan:
If you look at http://php.net/manual/fr/function.dirname.php you can see that dirname() can return "\" in some cases, so there is inevitably a bug if you don't merge the request.

NiR-:
Thansk @VanVan for the report. We have finally wrapped every `dirname` calls in order to only work with unix-style paths (see #397). Please, let us know if you have any other problems.

PR #349, opened 2015-07-31
Use getPartialObject for exists check.

jenkoian:
Means we simply get the head of the object to check it exists rather
than read the entire object which can take a long time if particularly
large file.

Ideally would have liked to have used the [objectExists](https://github.com/rackspace/php-opencloud/blob/working/lib/OpenCloud/ObjectStore/Resource/Container.php#L404) check but that's on a a newer version that Gaufrette currently doesn't support. If we upgrade the version of OpenCloud we should look to use that rather than getPartialObject but they both achieve the same outcome.


studioromeo:
:+1:

jenkoian:
Anyone available to take a look at this?

NiR-:
Thank you for the PR @jenkoian.

Issue #223, opened 2013-10-21
[RC] Amazon adapters should filter by directory when listing keys


alex88:
The current implementation of keys() in the `AmazonS3` adapter lists all keys of the bucket, since it has the directory information I think it should instead return only the keys inside that folder.
I discovered this since when we write files like `foo1`, `foo2` etc. and I want to download all the files beginning with `foo` I would do `listKeys('foo')` on the filesystem but this fails since the returned keys has the directory in the string.

What you think?

benoitMariaux:
:+1:

PR #351, opened 2015-09-10
List keys without options['directory'] on AwsS3 adapter

anacicconi:
This is a suggestion to correct this issue:

https://github.com/KnpLabs/Gaufrette/issues/223

The AwsS3 adapter was returning a list of keys with the options['directory'] in the path. This is not the behavior of other adapters (e.g: Local). And can be annoying if we try, for example, to duplicate a key on the bucket just by listing the keys and then writing them. The result will be a directory with the same name inside the main directory.

### DIFF
--- a/src/Gaufrette/Adapter/AwsS3.php
+++ b/src/Gaufrette/Adapter/AwsS3.php
@@ -291,6 +291,6 @@ class AwsS3 implements Adapter,
      */
     protected function computeKey($path)
     {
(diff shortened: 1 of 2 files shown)


benoitMariaux:
It works fine, thanks.
I did my own S3 adapter to implement your solution.

anacicconi:
I'm glad it helped @benoitMariaux

akovalyov:
@anacicconi thank you for the contribution. I will merge it right now and and it will be included in the nearest release.

PR #352, opened 2015-09-29
Correctly set cache control and other headers in Google Cloud Storage

watermanio:
The Google PHP API does not seem to use the metadata to set the permitted headers, only for custom data which is prepended with “x-goog-meta”. [See Docs](https://cloud.google.com/storage/docs/gsutil/addlhelp/WorkingWithObjectMetadata)

This fix sets the supported header fields directly onto the object class provided within the PHP API.

Usage example within the new test.


NiR-:
Thanks @watermanio for your PR.

PR #356, opened 2015-11-11
Fix duplicates file content when read file using MogileFS adapter

mikhalchuk:
This PR fixes duplicate file content when read file using MogileFS adapter.
GET_PATHS returns all urls to the same file for each tracker we configured as a replication.


NiR-:
Thanks.

PR #357, opened 2015-11-15
Add support for Backblaze B2 Cloud Storage

kamilZ:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Tests pass? | yes |
| Fixed tickets | no |
| License | MIT |

---
This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/357)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ tests/adapters/*
 !tests/adapters/*.dist
 bin/*
 !bin/configure_test_env.sh
+/.idea
+*.iml
--- a/.gitignore
+++ b/.gitignore
@@ -4,5 +4,3 @@ tests/adapters/*
 !tests/adapters/*.dist
 bin/*
 !bin/configure_test_env.sh
-/.idea
-*.iml
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,8 @@
         "dropbox-php/dropbox-php": "*",
         "herzult/php-ssh": "*",
         "phpunit/phpunit": "3.7.*",
-        "microsoft/windowsazure": "dev-master",
+        "microsoft/windowsazure": "*",
+        "kamilz/b2backblaze": "dev-master",
         "mikey179/vfsStream": "~1.2.0"
     },
     "suggest": {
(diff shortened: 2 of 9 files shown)


.gitignore
```diff
@@ -4,3 +4,5 @@ tests/adapters/*
 !tests/adapters/*.dist
 bin/*
 !bin/configure_test_env.sh
+/.idea
```
akovalyov: Could you remove these two lines?
I doubt that open-source libs should take care about particular IDE.

composer.json
```diff
@@ -1,71 +1,72 @@
 {
-    "name":         "knplabs/gaufrette",
```
akovalyov: Could you revert the formatting for better readability?

src/Gaufrette/Adapter/BackblazeB2Storage.php
```diff
@@ -0,0 +1,234 @@
...
+            return false;
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isDirectory($key)
+    {
+        if ($this->exists($key . '/')) {
```
akovalyov: I am not familiar with `Backblaze`, but what happens if I pass `/foo/bar/` directory as parameter here?

.gitignore
kamilZ: Ok

src/Gaufrette/Adapter/BackblazeB2Storage.php
kamilZ: Always we'll get bad request: "file names must not end with '/'". 
There is no way to check if "/foo/bar/" exist in B2 atm.

I'll change this to return false always.

composer.json
kamilZ: sorry for that.

NiR-:
Sorry @kamilZ for the really late response.

Unfortunately, we'll not merge your new adapter because we want to reduce the number of them we need to maintain. But, in the case you release your adapter as an independant FOSS project, I would be really happy to add a link in our readme/docs to your adapter :)

AntoineLelaisant:
Could we close this PR then @NiR- ?

PR #359, opened 2015-12-23
If a file we want to delete does not exist, we already deleted it successfully!

### LINKED ISSUE #318: Could not remove the "default/0001/01/70d43d74fb22538adbdc981eb6b6e3e4b318dca6.png" key.
Hi,

In my project we are using this bundle for some file uploading functionality, I dont have much knowledge about this, before it was working fine but I am facing this error. When I update new image then it works fine but when update same image with new upload again then this error come. I tried to solve it but unable to find the issue.

please help to resolve this issue.

Could not remove the "default/0001/01/70d43d74fb22538adbdc981eb6b6e3e4b318dca6.png" key.

I saw it gives exception from this file.
/vendor/knplabs/gaufrette/src/Gaufrette/Filesystem.php

```
/**
 * Deletes the file matching the specified key
 *
 * @param string $key
 * @throws \RuntimeException when cannot read file
 *
 * @return boolean
 */
public function delete($key)
{
    $this->assertHasFile($key);

    if ($this->adapter->delete($key)) {
        return true;
    }

    throw new \RuntimeException(sprintf('Could not remove the "%s" key.', $key));
}
```

snapshot332

lemoinem:
Fixes #318

### DIFF
--- a/spec/Gaufrette/FilesystemSpec.php
+++ b/spec/Gaufrette/FilesystemSpec.php
@@ -240,14 +240,11 @@ class FilesystemSpec extends ObjectBehavior
     /**
      * @param \Gaufrette\Adapter $adapter
      */
-    function it_does_not_delete_file_which_does_not_exist($adapter)
+    function it_should_consider_a_non_existent_file_as_deleted($adapter)
     {
         $adapter->exists('filename')->willReturn(false);
 
-        $this
-            ->shouldThrow(new \Gaufrette\Exception\FileNotFound('filename'))
-            ->duringDelete('filename')
-        ;
+        $this->delete('filename')->shouldReturn(true);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


lemoinem:
Travis with PHP 5.4 fails because the php-ext2 extension is not available.
This is not related to my changes.

NiR-:
@lemoinem I am not really confident with your PR: any filesystem would complain if you try to delete a file that does not exists rather than silently failing. I think we should keep the same behavior for our filesystem abstraction layer.

akovalyov:
I completely agree with @NiR-. If file was not found - it is unexpected behaviour. I.e., `rm` will through you `rm: cannot remove ‘my.file’: No such file or directory`. 
If such a behavior is expected - then the exception thrown by Gaufrette should be caught in your code.

Another question is that `\RuntimeException` is thrown there. It should be changed to `Gaufrette\Exception\FileNotFound`.

lemoinem:
OK, Thanks for the feedback...

PR #360, opened 2015-12-28
ObjectStoreFactory currently force 'cloudFiles' as Object Store type value

p-bizouard:
I need to use 'swift' as object store type and this change allow me to do it.


NiR-:
@p-bizouard Thanks for your PR but please can you clean it (squash commits, use a real commit message and author name) ?

PR #361, opened 2016-01-04
fix directory list local adapter

andrey1s:


tests/Gaufrette/Functional/Adapter/LocalTest.php
```diff
@@ -106,6 +107,37 @@ public function shouldListingOnlyGivenDirectory()
 
     /**
      * @test
+     * @covers Gaufrette\Adapter\Local
+     */
```
NiR-: I think you should also add `@group functional` here.

andrey1s: @NiR- `updated`

PR #365, opened 2016-01-16
Improve directory handling in StreamWrapper

### LINKED ISSUE #353: stat doesn't work with local adapter + stream wrapper
##### Problem

With regards to the local adapter, I simply want to be able to do something like is_dir('gaufrette://local/existing-directory/'); (or file_exists), but this will always return false.

##### Workarounds

The only simple workaround that I currently have is file_exists('gaufrette://local/existing-directory/existing-file.jpg'), but this isn't ideal in all cases, since I have to know about that file.

I can also $fileSystem->has('/existing-directory/'), but this is a lot more complicated as I have to fetch the filesystem via DI (am using knp-gaufrette-bundle) and strip off the protocol/FS from the path.

##### Cause

StreamWrapper->url_stat calls Stream\Local->open which fails as it attempts to fopen the directory and ends up throwing an exception. url_stat catches the exception and returns false.

Even if this were changed, methods in Stream\Local such as stat check for the initialization of $this->fileHandle which is set by the call to open and simply return false if not.

##### Comments

Local filesystems obviously support directory functionality, so I would expect calls to stat on a directory to succeed. I also found the initial addition of stat support in pull request #86, and it looks like this functionality would have worked at that time as Local's stat simply returns stat($this->path). Is this a regression or intended? Thanks! :)

NiR-:
Fix #353.

It is now possible to call `file_exists` & `is_dir` on stream URI pointing
to a directory, whether it is hosted locally or remotely.

### DIFF
--- a/spec/Gaufrette/StreamWrapperSpec.php
+++ b/spec/Gaufrette/StreamWrapperSpec.php
@@ -248,10 +248,12 @@ class StreamWrapperSpec extends ObjectBehavior
     /**
      * @param \Gaufrette\Stream $stream
      */
-    function it_does_not_stat_when_cannot_open($stream)
+    function it_stats_even_if_it_cannot_be_open($filesystem, $stream)
     {
+        $filesystem->createStream('dir/')->willReturn($stream);
         $stream->open(Argument::any())->willThrow(new \RuntimeException);
-        $this->url_stat('gaufrette://some/filename', STREAM_URL_STAT_LINK)->shouldReturn(false);
+        $stream->stat(Argument::any())->willReturn(array('mode' => 16893));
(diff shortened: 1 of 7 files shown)


spec/Gaufrette/StreamWrapperSpec.php
```diff
@@ -248,10 +248,12 @@ function it_should_stat_from_url($stream)
     /**
      * @param \Gaufrette\Stream $stream
      */
-    function it_does_not_stat_when_cannot_open($stream)
+    function it_stat_even_if_it_cannot_be_open($filesystem, $stream)
```
akovalyov: `it_stats`?

PR #366, opened 2016-01-17
On delete function, check to see if we are removing a folder(rmdir) o…

yunai39:
…r a file(unlink)


yunai39:
@akovalyov, Here is the pull request (Hope I did it right, never did this before)

NiR-:
Hey @yunai39, thank you for your PR but I'm sorry soon we will drop ssh extension support. I could only recommend you to use phpseclib sftp adapter.

PR #371, opened 2016-01-18
Fix phpsecsftp adapter

akovalyov:


stof:
@umpirsky merging a BC break (replacing the version 1 of phpseclib by the version 2) in a patch release without even documenting it in the release note is a bad experience for users of the library...

umpirsky:
@stof Nice catch, I agree. Sorry if it made any troubles for you.

PR #380, opened 2016-02-12
FIX High memory usage over time

robhunt3r:
FIX Memory problems when creating images from an iterator or any kind of script that constantly uses this method.

Without this fix, creating 1000 images with a 0.6mb size, ended with 700mb script in memory, but it get worse, memory size increases over time significantly.

This fix makes constant (or almost) the memory usage. Script started with 395MB in memory, 4000 iterations later it's still 395MB and images are created perfectly.

### DIFF
--- a/src/Gaufrette/File.php
+++ b/src/Gaufrette/File.php
@@ -141,8 +141,10 @@ class File
     {
         $this->content = $content;
         $this->setMetadata($metadata);
+        $this->size = $this->filesystem->write($this->key, $this->content, true);
 
-        return $this->size = $this->filesystem->write($this->key, $this->content, true);
+        $this->content = null;
+        return $this->size;
     }
 
     /**


NiR-:
Thanks for that PR, @robhunt3r. Dumb question for you: do you really need to keep your 1000 images in memory ?
In fact I'm not really convinced by your PR: it won't break the API but it would rather change how `File` operates, which means it would probably makes more call to remote services.

@docteurklein @akovalyov @umpirsky What do you think ?

Oh and btw, we are working on a new version of Gaufrette that will fix different high memory consumption flaws.

umpirsky:
For me it's odd to do `$file->setContent($content)` and have `$file->content` set to `null` as a result.

docteurklein:
as @umpirsky said, it breaks the file VO behaviour. not sure how we could do this

umpirsky:
@robhunt3r But in your code you can do:

```php
$size = $file->setContent($content);
$file->setContent(null);
```

or override `File`.

robhunt3r:
@NiR-  Well, I don't store 1000 images in memory, in fact I try to clean all my memory and arrays and stuff each iteration, but memory keeps increasing, it just stops when using what @umpirsky says, donig `$file->setContent(null);`, but I thought this PR would avoid doing that in our code, and let that class handle it, because until I knew where the problem comes from I spent a lot debugging, I thought it was my code.

NiR-:
@robhunt3r Are you using any cache adapter ? Did you try to call `$filesystem->removeFromRegister($key);` ?

robhunt3r:
Using @NiR- solution, removing from register clears the memory and everything works as spected, thank you.

Also, `$file->setcontent(null);` will end into writing 0bytes to the final file, because the `File` method, writes again to the same destination a `null` content, that's why this PR clears only memory, you can close this I think.

PR #386, opened 2016-03-03
Fix/travis

akovalyov:


umpirsky:
:+1:

.travis.yml
```diff
@@ -1,7 +1,12 @@
 language: php
 
+sudo: false
+
+cache:
+  directories:
+    - $HOME/.composer/cache
```
stof: I suggest caching only the ``files`` subfolder. See https://github.com/puli/manager/pull/9#discussion_r40499480 for the explanation

.travis.yml
```diff
@@ -10,10 +15,12 @@ php:
 matrix:
     allow_failures:
         - php: hhvm
+    fast_finish: true
 
 before_script:
-  - ./bin/configure_test_env.sh
-  - composer self-update && composer install --dev --prefer-source --no-interaction
+  - echo "extension = mongo.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
```
stof: this should be skipped on HHVM (and PHP 7 but it is not yet enabled) as mongo.so is not available there

.travis.yml
```diff
@@ -10,10 +15,12 @@ php:
...
     allow_failures:
         - php: hhvm
+    fast_finish: true
 
 before_script:
-  - ./bin/configure_test_env.sh
-  - composer self-update && composer install --dev --prefer-source --no-interaction
+  - echo "extension = mongo.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
+  - cp tests/Gaufrette/Functional/adapters/DoctrineDbal.php.dist tests/Gaufrette/Functional/adapters/DoctrineDbal.php -f
+  - composer self-update && composer install --dev --prefer-source --no-interaction --ignore-platform-reqs
```
stof: ``--dev``is a deprecated no-op option (well, actually, it does something: displaying a deprecation warning)

akovalyov:
@stof thank you for your valuable comments

PR #395, opened 2016-03-07
Update AwsS3.php

jjborie:
If there is an exception on put object it should be returned instead of masking it returning false.

---
This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/395)

### DIFF
--- a/src/Gaufrette/Adapter/AwsS3.php
+++ b/src/Gaufrette/Adapter/AwsS3.php
@@ -143,7 +143,7 @@ class AwsS3 implements Adapter,
 
             return strlen($content);
         } catch (\Exception $e) {
-            return false;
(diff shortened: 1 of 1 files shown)


akovalyov:
@jjborie thank you for reporting that. It is something I am working on at the moment in https://github.com/KnpLabs/Gaufrette/tree/feature/exceptions. You can take a look and advice me there if you notice anything.

NiR-:
Just realising this proposed patch is: first totally broken because

NiR-:
This patch is not ideal: the exception should be passed as previous and not to sprintf. Also I think you made an error because the stacktrace is printed in the middle of the sentence, that does not make sense.

Anyway, thank you for pointing out this issue to us. I've opened an issue to remind us to fix this bad error handling scheme for v1.0 (see #497).

Sorry for the first email with bad tone :/

PR #397, opened 2016-03-09
Wrap native dirname function in order to handle only UNIX-style paths

NiR-:
As stated by php doc (http://php.net/manual/en/function.dirname.php), dirname
can return either slash or back-slash as directory separator.

It tries to fix problems on windows in a more generic way than #347 and #271.

Note that I don't have any windows system in order to test. I think we should wait AppVeyor integration before going further.

### DIFF
--- a/src/Gaufrette/Adapter/Dropbox.php
+++ b/src/Gaufrette/Adapter/Dropbox.php
@@ -137,8 +137,8 @@ class Dropbox implements Adapter
         foreach ($metadata['contents'] as $value) {
             $file = ltrim($value['path'], '/');
             $keys[] = $file;
-            if ('.' !== dirname($file)) {
-                $keys[] = dirname($file);
+            if ('.' !== \Gaufrette\Util\Path::dirname($file)) {
+                $keys[] = \Gaufrette\Util\Path::dirname($file);
             }
         }
         sort($keys);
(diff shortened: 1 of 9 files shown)


src/Gaufrette/Adapter/Ftp.php
```diff
@@ -77,7 +77,7 @@ public function write($key, $content)
         $this->ensureDirectoryExists($this->directory, $this->create);
 
         $path = $this->computePath($key);
-        $directory = str_replace('\\', '/', dirname($path));
+        $directory = str_replace('\\', '/', \Gaufrette\Util\Path::dirname($path));
```
stof: no need to use ``str_replace`` here anymore, as ``\Gaufrette\Util\Path::dirname`` is doing it already

src/Gaufrette/Adapter/Dropbox.php
```diff
@@ -137,8 +137,8 @@ public function keys()
         foreach ($metadata['contents'] as $value) {
             $file = ltrim($value['path'], '/');
             $keys[] = $file;
-            if ('.' !== dirname($file)) {
-                $keys[] = dirname($file);
+            if ('.' !== \Gaufrette\Util\Path::dirname($file)) {
+                $keys[] = \Gaufrette\Util\Path::dirname($file);
```
stof: you should store it in a local variable to do the work only once

src/Gaufrette/Adapter/Ftp.php
stof: same in all other places doing this replacement btw

Issue #394, opened 2016-03-07
Cannot use phpseclib\Net\SFTP as SFTP because the name is already in use in ../knplabs/gaufrette/src/Gaufrette/Adapter/PhpseclibSftp.php on line 6


playr1983g:
I get the following error when using Gaufrette in a Symfony 2.4 environment when i do a cac:clear:

  `Cannot use phpseclib\Net\SFTP as SFTP because the name is already in use in ../knplabs/gaufrette/src/Gaufrette/Adapter/PhpseclibSftp.php on line 6`

I'm using the following versions:

knplabs/gaufrette | 0.2.1
knplabs/knp-gaufrette-bundle | 0.2.0

akovalyov:
We are talking about https://github.com/KnpLabs/Gaufrette/blob/0.2.1/src/Gaufrette/Adapter/PhpseclibSftp.php#L6, right?
Do you have `ssh2` extension installed?

playr1983g:
@akovalyov yes, that use statement.

I removed that **use** statement and  changed the constructor parameters from:

`public function __construct(SFTP $sftp,...)`

to 

`public function __construct(Sftp $sftp, ...)`

and it worked for me for the moment.
I have no ssh2 extension installed. My PHP version is PHP 5.5.9-1ubuntu4.14
I also tried to install the ssh2 extension - if needed - with no effect..

http://www.codexpedia.com/apache-server/install-phps-ssh2-extension-on-ubuntu-and-mac/

akovalyov:
@playr1983g thanks for your report.

well, I am not quite sure from where does this issue come. Which version of `phpseclib` do you have? Could you list your PHP modules with `php -m`?

Thanks in advance.

akovalyov:
@playr1983g just forgot to clarify - no, `ssh2` extension is not needed. I was just asking that because according to your error message, there is already a global class with `SFTP` name and you have a name collision there, so I just was thinking about that.
PHPSecLib means pure php impl of several protocols, there is no need of ssh2 extension there.

playr1983g:
Ok thanks @akovalyov 

phpseclib:

`phpseclib/phpseclib                  2.0.1`

This my `php -m` output  (now of course with ssh2 included because i installed it :) ) 

```
[PHP Modules]
apc
apcu
bcmath
bz2
calendar
Core
ctype
curl
date
dba
dom
ereg
exif
fileinfo
filter
ftp
gd
gettext
hash
iconv
intl
json
[...]

PR #399, opened 2016-03-22
Fix for "name already in use"

RyanMcEvoy:
Fix for "Cannot use phpseclib\Net\SFTP as SFTP because the name is already in use" when a compiler pass is in use on a Symfony Framework.

Solves #394


stof:
:+1:

akovalyov:
thanks @RyanMcEvoy

stof:
@akovalyov could you tag a release containing this bugfix ?

vicdelfant:
Any progress on this? Right now we have to manually require the `0.4.*@dev` package in our production environments, which isn't ideal. Would be great if this could be tagged as a `0.2.x` release.

PR #406, opened 2016-04-15
Added the filesystem accessor to a File class.

robertpustulka:
Turns out I need to access the file's filesystem. So there's an accessor. Would be happy if you merged it :)

---
This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/406)

### DIFF
--- a/src/Gaufrette/File.php
+++ b/src/Gaufrette/File.php
@@ -89,6 +89,16 @@ class File
     }
 
     /**
+     * Returns Filesystem instance.
+     *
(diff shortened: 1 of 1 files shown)


NiR-:
@robertpustulka I'm not sure about your PR. Can you explain your use case?

robertpustulka:
I wanted to detect file source while working with many File objects coming from different filesystems.

Later I dropped that idea in my app, so I don't need that accessor, but you could still merge it and make it useful for somebody.

PR #411, opened 2016-05-20
Updated new Azure namespaces.

nico-incubiq:
Microsoft has changed the namespaces in the latest version of the azure-storage package (which has been split out from the SDK).
I updated the namespaces throughout the Gaufrette package, and also updated composer.json to add the conflict with previous versions of the azure SDK.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,6 +25,9 @@
     "require": {
         "php": ">=5.4"
     },
+    "conflict": {
+        "microsoft/windowsazure": "<0.4.3"
+    },
     "require-dev": {
         "aws/aws-sdk-php": "~2",
         "amazonwebservices/aws-sdk-for-php": "1.5.*",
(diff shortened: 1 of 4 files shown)


NiR-:
@akovalyov IMHO we should postpone this PR and merge it only after releasing the next version. What do you think?

nico-incubiq:
Hi,
I think you may want to include this fix as soon as possible, since the azure API itself has been updated and requires the latest version of the azure framework, which already uses the new namespaces.
For now I am using my forked version on prod with no issue.
Regards,
Nicolas

charlesastwood:
Will this be released soon as Azure Storage currently not working

mark-sp:
Please include this fix soon. It's just a minor change in the namespaces to fit the new specification by Microsoft:

To change in AzureBlobStorage:
use MicrosoftAzure\Storage\Blob\Models\CreateBlobOptions;
use MicrosoftAzure\Storage\Blob\Models\CreateContainerOptions;
use MicrosoftAzure\Storage\Blob\Models\DeleteContainerOptions;
use MicrosoftAzure\Storage\Blob\Models\ListBlobsOptions;
use WindowsAzure\Common\ServiceException;

nico-incubiq:
Hi,
This is no longer a BC break but rather a mandatory fix.
The current plugin does not work anymore with Azure and requires that this pull request to be merged.

PR #413, opened 2016-05-25
Remove useless str_replace

NiR-:


src/Gaufrette/Adapter/Dropbox.php
```diff
@@ -137,8 +137,8 @@ public function keys()
         foreach ($metadata['contents'] as $value) {
             $file = ltrim($value['path'], '/');
             $keys[] = $file;
-            if ('.' !== \Gaufrette\Util\Path::dirname($file)) {
-                $keys[] = \Gaufrette\Util\Path::dirname($file);
+            if ('.' !== $dirname = \Gaufrette\Util\Path::dirname($file)) {
```
akovalyov: Unnecessary trailing slash here and below in diff.

NiR-: What do you mean?

akovalyov: According to http://php.net/manual/en/language.namespaces.importing.php, leading backslash is not required. Moreover, it is not recommended. So, if you could just remove these backslashes - would be perfect. I agree that it might be out of scope of this PR, but since you are working with this diff, I hope it won't take much time for you to do that.

NiR-: @akovalyov Still not convinced: from my understanding leading backslashes are unneeded and not recommended in the case of import (`use`) since imported namespaces are always fully qualified.
Do you mean I should import the function ?

PR #419, opened 2016-06-26
Add a feature to get the name of the AWS Bucket

edhgoose:
This simply adds a way of exposing the bucket name from the adapter. This would be quite beneficial for me!

---
This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/419)


NiR-:
This adapter has been deprecated in favor of newer `AwsS3` adapter (see #464). Thank you for taking time to create this patch.

Issue #420, opened 2016-07-06
AWS S3 List Keys should add filesize and mtime to avoid roundtrip performance issue.


vincentwongso:
AWS S3 Adapter listKeys function should add some extra metadata like filesize and mtime to avoid roundtrip performance issues when we need to list all files in a directory. 

**Problem**
At the moment the only way to do this is by listing all keys and then loop through each key and call size() and mtime() function which means calling amazon api to get the details. This might be find if we are listing small amount of files but it will have huge performance impact when listing lots of files. 

**Proposed solution:**
Either add extra metadata(filesize and mtime) on listKeys which already returned by ListObjects (s3 api) anyway or create a new function to do this.

vincentwongso:
I have created a new function to avoid breaking the existing functionality, pull request:
https://github.com/KnpLabs/Gaufrette/pull/421

PR #421, opened 2016-07-06
Added listFiles function in AwsS3 adapter

vincentwongso:
Added listFiles function in AwsS3 adapter for ticket [#420](https://github.com/KnpLabs/Gaufrette/issues/420)

---
This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/421)

### DIFF
--- a/src/Gaufrette/Adapter/AwsS3.php
+++ b/src/Gaufrette/Adapter/AwsS3.php
@@ -221,6 +221,41 @@ class AwsS3 implements Adapter,
     }
 
     /**
+     * List files beginning with given prefix.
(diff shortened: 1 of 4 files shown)


vincentwongso:
Failed... all of the AwsS3Test functions are failing at the moment so it's not just my test. I think it might be related to the update on AwsS3 php library.

NiR-:
First of all, thank you for your interest for Gaufrette and sorry for the long delay.

I would prefer to find a proper and common way to handle metadatas across adapters rather than implementing another new interface, introducing even more differences between adapters and killing its main purpose. I created a note in our gh project to remember to do it soon.

Again thank you :)

PR #427, opened 2016-09-09
Adds adapter for Google Cloud Client Library 

PanzerLlama:
Hi,

Adapter for http://googlecloudplatform.github.io/google-cloud-php/

---
This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/427)


NiR-:
Hello @PanzerLlama,

What's the difference with the current [`GoogleCloudStorage`](https://github.com/KnpLabs/Gaufrette/blob/master/src/Gaufrette/Adapter/GoogleCloudStorage.php)?

PanzerLlama:
Hi,

This is basically a different library by Google for their cloud and a (better) alternative to GCS. Quote from the homepage (https://googlecloudplatform.github.io/google-cloud-php/#/):

"What is the relationship between the Google Cloud Client Library and the Google APIs PHP Client?

[...] The Google Cloud Client Library is built specifically for the Google Cloud Platform and is the recommended way to integrate Google Cloud APIs into your PHP applications. If your application requires both Google Cloud Platform and other Google APIs, the 2 libraries may be used by your application."

Personally I found the documentation for GCCL to be clearer and easier to work with, also this lib is actively developed by Google programmers contrary to the GCS which is generated automatically. In a nutshell - Google recommends GCLL if the app is developed in single language (PHP).

More details on the differences can be found here: https://cloud.google.com/storage/docs/json_api/v1/libraries

NiR-:
@PanzerLlama Could you update the existing adapter rather than creating a new one?

PanzerLlama:
@NiR- - not sure if this works out well since these are 2 different libraries sitting on top of the Google Cloud API so adding support to the existing adapter would generate a lot of extra code in it. Also the constructor of the existing adapter requires \Google_Service_Storage and my adapter uses \Google\Cloud\Storage\StorageClient so looks like merging will create BC break. (?) I think since these are 2 different libraries it feels right to have 2 different adapters.

composer.json
```diff
@@ -38,7 +38,8 @@
         "phpunit/phpunit": "3.7.*",
         "microsoft/windowsazure": "dev-master",
         "mikey179/vfsStream": "~1.2.0",
-        "league/flysystem": "~1.0"
+        "league/flysystem": "~1.0",
+        "google/cloud": ">=0.20.1"
```
PanzerLlama: Agreed & fixed.

AntoineLelaisant:
Hi @PanzerLlama Sorry for not responding for a long time but we currently having a big internal discussion about the future of Gaufrette (see https://github.com/KnpLabs/Gaufrette/issues/466 and https://github.com/KnpLabs/Gaufrette/issues/464). The main idea for now is to split adapters into meta-packages and limit the amount of adapters we have to maintain.

At the moment we already have a `GoogleCloudStorage` adapter based on https://github.com/google/google-api-php-client and we don't want to maintain two adapters for the same functionality at the end. Moreover we could not simply replace the `GoogleCloudStorage` adapter by yours because it would bring BC breaks.

By the way, your implementation seems cool and it would be a waste to simply throw it away. 

So I see two solutions:

1) Let this PR pending until the release of Gaufrette v1.0
    -> Deprecate the actual `GoogleCloudStorage` in Gaufrette v0.4
    -> Replace it by your adapter in v1.0

2) Create a separated repo for your adapter (so not maintained by KnpLabs)
We would be happy to mention it in the Gaufrette's README.md

NiR-:
@PanzerLlama Indeed, seems to be a bad idea to reuse the current Adapter.

@AntoineLelaisant I would go for the first option, but without even waiting the v1.0 release. If we want to deprecate the old adapter in the v0.4 release we need this new adapter to be available.

PanzerLlama:
Hi @AntoineLelaisant - let's do what you think is best. I'll be away now for around a week but will follow the feedback.

tests/Gaufrette/Functional/adapters/GoogleCloudClientStorage.php.dist
```diff
@@ -0,0 +1,19 @@
+<?php
```
NiR-: We're avoiding this way of running test. Could you update following the new recommended way (see #465)?

src/Gaufrette/Adapter/GoogleCloudClientStorage.php
```diff
@@ -0,0 +1,340 @@
...
+    public function setOptions($options)
+    {
+        $this->options = array_replace($this->options, $options);
+    }
+    
+    protected function computePath($key)
+    {
+        if (strlen($this->options['directory']))
+        {
+            if (strcmp(substr($this->options['directory'], -1), '/') == 0)
```
NiR-: Would be better to trim `directory` in the constructor.

src/Gaufrette/Adapter/GoogleCloudClientStorage.php
```diff
@@ -0,0 +1,340 @@
...
+            {
+                return $this->options['directory'].$key;
+            }
+            return $this->options['directory'].'/'.$key;
+        }
+        return $key;
```
NiR-: Is it possible to use `/` as a key? If that's possible and you if fix previous comment, you could factor `computePath` into a one-line method :)

composer.json
```diff
@@ -41,7 +41,8 @@
         "phpunit/phpunit": "3.7.*",
         "mikey179/vfsStream": "~1.2.0",
         "league/flysystem": "~1.0",
-        "mongodb/mongodb": "~1.1.4"
+        "mongodb/mongodb": "~1.1.4",
+        "google/cloud-storage": "dev-master"
```
NiR-: `google/cloud-storage` v1.0 was released a month ago. Do you think your work is compatible with this branch? Would be great to depend on a stable branch rather than `dev-master` ;)

doc/adapters/google-cloud-client-storage.md
```diff
@@ -0,0 +1,52 @@
...
+    array(
+        'FileDescription' => 'This is my file. There are many like it, but this one is mine.'
+    )
+);
+
+$filesystem = new Filesystem($adapter);
+
+$filesystem->write($key, 'Uploaded at: '.date('Y-m-d @ H:i:s'), true);
+
+```
```
NiR-: Do you have any idea what's the minimum set of security access (roles?) needed to use it? Would be awesome to provide something equivalent to [this](https://knplabs.github.io/Gaufrette/adapters/aws-s3.html#iam-policy). But that's a detail, it could be postpone to another PR.

src/Gaufrette/Adapter/GoogleCloudClientStorage.php
```diff
@@ -0,0 +1,340 @@
...
+    }
+    
+    /**
+     * {@inheritdoc}
+     */
+    public function read($key)
+    {   
+        $this->isBucket();     
+        $object = $this->bucket->object($this->computePath($key));
+        $info = $object->info();
```
NiR-: There's no error handling at all here.

src/Gaufrette/Adapter/GoogleCloudClientStorage.php
```diff
@@ -0,0 +1,340 @@
...
+    }
+    
+    /**
+     * {@inheritdoc}
+     */
+    public function exists($key)
+    {
+        $this->isBucket();
+        $object = $this->bucket->object($this->computePath($key));
+        if ($object->exists())
```
NiR-: You can `return $object->exists()` ;)

src/Gaufrette/Adapter/GoogleCloudClientStorage.php
```diff
@@ -0,0 +1,340 @@
...
+        }
+        return false;
+    }
+    
+    /**
+     * {@inheritdoc}
+     */
+    public function isDirectory($key)
+    {
+        if ($this->exists($key . '/'))
```
NiR-: Same here

src/Gaufrette/Adapter/GoogleCloudClientStorage.php
```diff
@@ -0,0 +1,340 @@
...
+    }
+    
+    /**
+     * {@inheritdoc}
+     */
+    public function listKeys($prefix = null)
+    {
+        $this->isBucket();        
+        $keys = array();        
+        if ($prefix === null)
```
NiR-: Why don't you allow null value for `computePath` so you won't have two ways of initializing `$prefix`? Btw, it's considered as a bad practice to initialize variables in if/else without default value outside: variables should be defined, at most, at the same level than they are used.

src/Gaufrette/Adapter/GoogleCloudClientStorage.php
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function mtime($key)
+    {
+        $this->isBucket();
+        $object = $object = $this->bucket->object($this->computePath($key));
+        $info = $object->info();
```
NiR-: No error handling here too, is it normal?

src/Gaufrette/Adapter/GoogleCloudClientStorage.php
```diff
@@ -0,0 +1,340 @@
...
+    }
+    
+    /**
+     * {@inheritdoc}
+     */
+    public function delete($key)
+    {
+        $this->isBucket();
+        $object = $this->bucket->object($this->computePath($key));
+        $object->delete();
```
NiR-: Same question as previous comment.

src/Gaufrette/Adapter/GoogleCloudClientStorage.php
```diff
@@ -0,0 +1,340 @@
...
+    protected function updateKeyProperties($key, $options = array())
+    {
+        if ($this->exists($key) === false)
+        {
+            return false;
+        }
+        
+        $object = $this->bucket->object($this->computePath($key));
+        
+        $properties = array_replace_recursive(
```
NiR-: Why `replace_recursive`? A simple `array_replace` should do the work, no?

src/Gaufrette/Adapter/ResourcesSupporter.php
```diff
@@ -0,0 +1,28 @@
+<?php
+namespace Gaufrette\Adapter;
+
+/**
+ * Interface for storing the resource information
+ *
+ * @author  Lech Buszczynski <lecho@phatcat.eu>
+ */
+interface ResourcesSupporter
```
NiR-: I'm not sure about this one? What's the purpose of such interface?

PanzerLlama:
@NiR- Will do the fixes sometime next week. Thanks for the feedback.

src/Gaufrette/Adapter/GoogleCloudClientStorage.php
PanzerLlama: / gives 404 (just a note)

NiR-:
@PanzerLlama Is it ready to be merged?

PanzerLlama:
@NiR- yes, done all requested changed (may improve docs later but for now I am done)

bwegrzyn:
Can't wait for this adapter to be merged. 👍 

Thanks for the hard work.

Matan:
Hi! Any update on this? Would be great if we could get this merged in. :+1:

PanzerLlama:
@NiR-  I am not sure why this still is labelled "needs rework" - I am happy to do changes but as far as I can see I did the updates that were requested.

PanzerLlama:
For anyone who wants to use this adapter (workaround): 

1) copy this file into your adapters directory: https://github.com/PanzerLlama/Gaufrette/blob/master/src/Gaufrette/Adapter/GoogleCloudClientStorage.php

2) read: https://github.com/PanzerLlama/Gaufrette/blob/master/doc/adapters/google-cloud-client-storage.md

You can also copy the test but its not required. I use this adapter for well over a year w/o issues.

Issue #401, opened 2016-03-24
Update Google Cloud Storage adapter


ivanjaros:
So I have been trying to setup GCS integration and found some issues. Mainly I cannot read files, though I am able to create them. I have found the issue is with calling the old Http classes that were completely removed in favour of Guzzle(like Google_Http_Request in \Gaufrette\Adapter\GoogleCloudStorage::read). So currently this is not usable for me at all.

https://github.com/google/google-api-php-client/blob/master/UPGRADING.md

cdomigan:
Is this still an issue? Deciding whether to use this library in a project with a GCS backend.

ivanjaros:
I have abandoned GCS in favour of S3(even though I'm still on GCE). But I'll be having a second look at this soon. There another composer project that should be supporting GCS and is more updated than gaufrette but I don't remember the name right now.

PanzerLlama:
I did not have any problems with this when using that adapter recently.

carlos-granados:
It only works with version 1.x of the Google Cloud php API, does not work with version 2.x. Both versions should be supported

carlos-granados:
I have created a PR to solve this https://github.com/KnpLabs/Gaufrette/pull/429

PR #429, opened 2016-09-20
Modify Google Cloud Storage Adapter

carlos-granados:
to work with version 2.x of the Google APIs Client Library for PHP


carlos-granados:
Solves https://github.com/KnpLabs/Gaufrette/issues/401

architech99:
Any chance this fix will get merged in the near future? If I copy the contents of the GoogleCloudStorage class file into my local vendor folder, the error I get from a missing Google_Http_Request class goes away and everything works fine. It's looking like the issue is with PHP 5.4 and 5.5 builds, so I'm not sure what (if any) priority there is to address this.

PR #433, opened 2016-11-16
[WIP] Move Local adapter into his own namespace

NiR-:


src/Gaufrette/Adapter/Local/.travis.yml
```diff
@@ -0,0 +1,23 @@
...
+matrix:
+  allow_failures:
+    - php: hhvm
+
+before_script:
+  - composer install --prefer-source --no-interaction
+
+script:
+  - vendor/bin/phpspec run -fpretty --verbose
+  - vendor/bin/phpunit
```
stof: Why having a Travis config in the subfolder ?

src/Gaufrette/Adapter/Local/composer.json
```diff
@@ -0,0 +1,31 @@
+{
+    "name": "gaufrette/local-adapter",
```
stof: if you also want to expose it as its own package, you must have the replace rule in the main composer.json too, otherwise things will behave weird.

And providing subtree splits for adapters does not make any sense if the core does not have its own subtree split as well. The adapter package must not depend on the full repo (as the full repo also contains the adapter code)

PR #435, opened 2016-12-14
Move AwsS3 adapter into his own namespace

NiR-:


composer.json
```diff
@@ -29,7 +29,6 @@
         "microsoft/windowsazure": "<0.4.3"
     },
     "require-dev": {
-        "aws/aws-sdk-php": "~2",
```
stof: should not be removed, otherwise tests are not running anymore

stof:
@NiR- as said previously, splitting adapters makes sense only if they don't depend on the main gaufrette package (as the main package also contain the adapters and so must rpelace the subtree splits). This implies that adapters should not live in the same tree than the core. Currently, you still put them in a subfolder of the core, making it impossible to split things cleanly.

NiR-:
@stof It's not done yet but it doesn't mean it's not in the pipes.

PR #440, opened 2016-12-14
Split adapters

NiR-:
Split adapters
==========

https://php-and-symfony.matthiasnoback.nl/2014/04/theres-no-such-thing-as-an-optional-dependency/#the-solution

Why?
-------

Currently, maintaining the whole adapters suite is tough:
* from time to time, tests break on some adapters and PRs (on another adapter) just can't be merged
* there're some risks of conflicting requirements between underlying SDKs (already happened! #426)
* some adapters should be deprecated and might be replaced by something else (e.g. `Cache` adapter)

Moreover there are some architectural problems:
* Big files are not properly handled (but #334 might be a good start)
* [Metadata are not properly handled too](https://github.com/KnpLabs/Gaufrette/issues/108)
* Weird and/or duplicated behavior: `Cache` adapter & `fileRegister` in filesystem.
* Sometimes connection to external service is handled in the Adapter (e.g. [FTP](https://github.com/KnpLabs/Gaufrette/blob/master/src/Gaufrette/Adapter/Ftp.php#L499-L550)), sometimes it is not (e.g. [AwsS3](https://github.com/KnpLabs/Gaufrette/blob/master/src/Gaufrette/Adapter/AwsS3.php#L27))
* Lack of proper exceptions #395

How?
-------

Split adapters across multiple repositories, then operate needed architectural changes.

TODO:
* [ ] Deprecate adapters we don't want to support anymore
  * [ ] AmazonS3
  * [ ] Cache *(should be rewritten with PSR-16 in mind)*
* [ ] Split every adapter in its own namespace and merge it in this branch,
  * [ ] AwsS3
  * [ ] AzureBlobStorage
  * [ ] DoctrineDbal
  * [ ] Dropbox
  * [ ] Flysystem
  * [ ] Ftp
  * [ ] GoogleCloudStorage
  * [ ] GridFS
  * [ ] InMemory
  * [ ] Local
  * [ ] MogileFS
  * [ ] OpenCloud
  * [ ] PhpseclibSftp
  * [ ] Sftp
  * [ ] Zip
* [ ] Merge every possible PRs on master, release the last version on this repository, and freeze the repository,
* [ ] Merge modifications from master into this branch,
* [ ] Push the core in its new repository, update every composer.json,
* [ ] Push every adapter in its new repository
* [ ] Plans for needed architectural changes

PR #441, opened 2016-12-15
Move Phpseclib adapter into its own namespace

NiR-:
See #440


src/Gaufrette/Adapter/PhpseclibSftp.php
```diff
@@ -2,11 +2,16 @@
...
+
 use Gaufrette\Adapter;
 use phpseclib\Net\SFTP as SecLibSFTP;
 use Gaufrette\Filesystem;
 use Gaufrette\File;
 
+/**
+ * @deprecated 0.4 This adapter is deprecated since version 0.4. Use Gaufrette\Adapters\Phpseclib\Sftp instead.
+ */
 class PhpseclibSftp implements Adapter,
```
PedroTroller: I think the old classe should extends from the new one to not duplicates the code.

src/Gaufrette/Adapter/Phpseclib/composer.json
```diff
@@ -0,0 +1,30 @@
...
+    "name": "gaufrette/phpspeclib-adapter",
+    "type": "library",
+    "require": {
+        "php": ">=5.6",
+        "phpseclib/phpseclib": "^2.0",
+        "knplabs/gaufrette": "0.4.*@dev"
+    },
+    "require-dev": {
+        "phpspec/phpspec": "^3.1",
+        "phpunit/phpunit": "3.7.*"
```
stof: 3.7 ? Seriously ? This is unmaintained since 5 years at least.

Use ``^4.8.35 || ^5.7`` instead (and then work on switching to the namespaced test case and add `` || ^6`` in the condition)

src/Gaufrette/Adapter/Phpseclib/phpunit.xml.dist
```diff
@@ -0,0 +1,31 @@
...
+<phpunit backupGlobals="false"
+         backupStaticAttributes="false"
+         colors="true"
+         convertErrorsToExceptions="true"
+         convertNoticesToExceptions="true"
+         convertWarningsToExceptions="true"
+         processIsolation="false"
+         stopOnFailure="false"
+         syntaxCheck="false"
+         bootstrap="tests/bootstrap.php.dist"
```
stof: why .dist ?

src/Gaufrette/Adapter/Phpseclib/tests/bootstrap.php.dist
```diff
@@ -0,0 +1,7 @@
+<?php
+
+$baseDir = dirname(__DIR__);
+
+$loader = require __DIR__.'/../vendor/autoload.php';
+$loader->add('Gaufrette', array($baseDir.'/vendor/knplabs/gaufrette/tests/'));
```
stof: use ``autoload-dev`` instead of altering the configuration of the generated loader. this will make this whole file useless

src/Gaufrette/Adapter/Phpseclib/tests/bootstrap.php.dist
```diff
@@ -0,0 +1,7 @@
...
+$loader->register();
```
stof: useless as it is already registered in ``autoload.php``

src/Gaufrette/Adapter/PhpseclibSftp.php
```diff
@@ -2,11 +2,16 @@
 
 namespace Gaufrette\Adapter;
 
+@trigger_error('The '.__NAMESPACE__.'\PhpseclibSftp is deprecated since version 0.4. Use Gaufrette\Adapters\Phpseclib\Sftp instead.', E_USER_DEPRECATED);
```
stof: wrong namespace. ``Adapter`` is not a plural

Issue #442, opened 2016-12-27
Checksum using other algorithm


IndraGunawan:
current checksum util only using md5. how about adding checksum using algorithm, like sha1 or sha256?

PR #443, opened 2016-12-28
add algo to checksum

IndraGunawan:
#442

---
This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/443)

### DIFF
--- a/src/Gaufrette/Adapter/ChecksumCalculator.php
+++ b/src/Gaufrette/Adapter/ChecksumCalculator.php
@@ -10,11 +10,12 @@ namespace Gaufrette\Adapter;
 interface ChecksumCalculator
 {
     /**
-     * Returns the checksum of the specified key.
+     * Returns the checksum of the specified key and algorithm.
      *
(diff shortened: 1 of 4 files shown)


src/Gaufrette/Adapter/ChecksumCalculator.php
```diff
@@ -10,11 +10,12 @@
...
-     * Returns the checksum of the specified key.
+     * Returns the checksum of the specified key and algorithm.
      *
      * @param string $key
+     * @param string $algo
      *
      * @return string
      */
-    public function checksum($key);
+    public function checksum($key, $algo);
```
NiR-: * Unfortunately, excluding `Local` adapter, there are many other places where the change needs to be done
* It would be a BC break for anyone having an adapter outside of this project
* It brings a new dependency, `hash` ext

I'm not against the idea. But I don't think we should do this right now, on `master`. It might better suite the refacto prepared on `split-adapters` branch.

PR #444, opened 2017-01-17
Timeout Option for FTP Configuration

KonstantinCodes:
FTP Timeouts caused our Nginx to produce `504 Gateway Time-out` Errors.
Here you can set a timeout configuration that plays nice with your individual setup.
In Collaboration with @cmyker

### DIFF
--- a/src/Gaufrette/Adapter/Ftp.php
+++ b/src/Gaufrette/Adapter/Ftp.php
@@ -25,6 +25,7 @@ class Ftp implements Adapter,
     protected $create;
     protected $mode;
     protected $ssl;
+    protected $timeout;
     protected $fileData = array();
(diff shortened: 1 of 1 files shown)


mvhirsch:
I would love to see this feature soon too! So I want to offer help.

@flug I just checked the failed builds, but it's about parsing error in Doctrine's Collection class. This doesn't look like an issue by @KonstantinCodes' changes, does it? Can you please take a look?

flug:
Hum maybe we need to change the version of doctrine/dbal. 
@NiR- an idea or it's good ?

mvhirsch:
@flug @NiR- any news so far?

mvhirsch:
Thank you @NiR-

PR #445, opened 2017-02-16
Ftp Adapter: Use „Create Dir“ Parameter for Rename

KonstantinCodes:
When I used the Rename Operation, I noticed that the SFTP Adapter automatically creates the new Directory if it doesn't exist, if the **create** option is `true`.

At the moment, FTP Adapter just throws an exception when the destination dir doesn't exist, even if the **create** flag is set.

This change basically makes the FTP Adapter rename like the SFTP Adapter

---
This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/445)

### DIFF
--- a/src/Gaufrette/Adapter/Ftp.php
+++ b/src/Gaufrette/Adapter/Ftp.php
@@ -108,7 +108,7 @@ class Ftp implements Adapter,
         $sourcePath = $this->computePath($sourceKey);
         $targetPath = $this->computePath($targetKey);
 
-        $this->ensureDirectoryExists(\Gaufrette\Util\Path::dirname($targetPath));
+        $this->ensureDirectoryExists(\Gaufrette\Util\Path::dirname($targetPath), $this->create);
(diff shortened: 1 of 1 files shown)


src/Gaufrette/Adapter/Ftp.php
```diff
@@ -108,7 +108,7 @@ public function rename($sourceKey, $targetKey)
         $sourcePath = $this->computePath($sourceKey);
         $targetPath = $this->computePath($targetKey);
 
-        $this->ensureDirectoryExists(\Gaufrette\Util\Path::dirname($targetPath));
+        $this->ensureDirectoryExists(\Gaufrette\Util\Path::dirname($targetPath), $this->create);
```
NiR-: It would be better to enfore the creation of $targetPath. `$this->create` is used to ensure the base directory exists.

NiR-:
Closed by #476.

PR #333, opened 2015-03-27
Adds mimeType to the AwsS3 adapter.

squaresurf:

Issue #274, opened 2014-03-28
Not implemented fully metadata support on S3


thekandor:
When we create object on S3 as one of option we can add only few headers which will be added as a metadata without conversion with 'x-amz-meta-' etc. Current version support only Content-Type, Content-Length. Method update_object from SDK isn't also used. But there are situations when that behaviour is needed, like setting object metadata which force Content-Disposition:attachment. Currently this is impossible. And both, old and new, versions of SDK allow for that in some cases. This may be even solved by adding those cases like Content-Type were solved. This bug is especially painfull, because IE in some cases (docx and this family) have problem with proper MimeType recognize (treat them as zip files).

darkbluesun:
Any update on this? It's a real pain.

barat:
I think that only way is to fork it and add own method to set Content-Type header :(

gggeek:
A sub-issue of https://github.com/KnpLabs/Gaufrette/issues/108 ?

Issue #108, opened 2012-11-21
Flawed metadata support


stof:
In the Amazon adapter, metadata are used from memory when renaming a file, but the metadata are never populated from the persistent ones.
The same issue exists for ``getMetadata``: it will return them only in the process defining them as it won't read them from amazon

wysow:
Closing this as it seems outdated now...

stof:
@wysow this is not outdated. The metadata handling has not been changed since 2012 and is still flawed.

NiR-:
Wow.. I wasn't aware of this issue until @zyphlar created #451. I will try to deep dive into @TomiS's branch and see what needs extrawork.

TomiS:
@NiR- Yeah, unfortunately my refactoring attempts are pretty much dead code after all these years :)

jspizziri:
@NiR- , did https://github.com/KnpLabs/Gaufrette/pull/451 resolve & close this issue?

grempa:
Are there any new updates to this issue?

PR #451, opened 2017-03-10
Add ContentType support to AwsS3

zyphlar:
Seems like there's a bug where metadata is never actually read from AWS.

Similar to / partially addresses #274 and #333

### DIFF
--- a/src/Gaufrette/Adapter/AwsS3.php
+++ b/src/Gaufrette/Adapter/AwsS3.php
@@ -92,8 +92,15 @@ class AwsS3 implements Adapter,
         $this->ensureBucketExists();
         $options = $this->getOptions($key);
 
+        // Copy ContentType metadata to local metadata
+        $object = $this->service->getObject($options);
+        if (!array_key_exists($key, $this->metadata) || !is_array($this->metadata[$key])) {
+            $this->metadata[$key] = array();
+        }
+        $this->metadata[$key]['ContentType'] = $object->get('ContentType');
+
         try {
-            return (string) $this->service->getObject($options)->get('Body');
+            return (string) $object->get('Body');
         } catch (\Exception $e) {
             return false;
         }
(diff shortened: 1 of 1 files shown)


NiR-:
Hello @zyphlar,

Thanks to bring us attention on this issue. As mentioned by @gggreek in #274, all these issues related to metadata (and especially Content-Type) are affiliated to #108. This issue indicates a systemic problem with metadata, hence it needs some architectural changes to be properly fixed.

So I will not merge your PR, but you can still subscribe/contribute on #108. As I wasn't aware of this issue until now, again thank you :)

zyphlar:
Thanks @NiR but I'm unclear on what the ultimate fix will be. Reading through the discussion it sounds like a lot of problems are raised without a real answer. What kind of architectural changes should be made? If I was going to contribute, where would that contribution be needed and what form would it take?

src/Gaufrette/Adapter/AwsS3.php
```diff
@@ -92,8 +92,15 @@ public function read($key)
         $this->ensureBucketExists();
         $options = $this->getOptions($key);
 
+        // Copy ContentType metadata to local metadata
+        $object = $this->service->getObject($options);
```
NiR-: Could you instead wrap this line into the try/catch block, rather than the `$object->get('Body')`.

src/Gaufrette/Adapter/AwsS3.php
```diff
@@ -92,8 +92,15 @@ public function read($key)
...
+        if (!array_key_exists($key, $this->metadata) || !is_array($this->metadata[$key])) {
+            $this->metadata[$key] = array();
```
NiR-: PHP5.3 is not supported anymore, you can use the short syntax.

NiR-:
Sorry for the delay. Can you update your PR as requested?

zyphlar:
@NiR- done

PR #452, opened 2017-03-10
Touch cached file mod time once every $ttl interval

zyphlar:
As currently written, it seems like `(time() - $this->ttl >= $dateCache)` will return true for all values of `time()` that are `ttl` seconds after the cached file was created. This means that `source->mtime` will be queried for basically every file request (except the requests in the first `ttl` seconds after the cache is created.)

If the `source` is indeed laggy, you'd think it would take a long time to query `source->mtime`. So, the ttl feature should avoid checking the source except once per `ttl` interval.

I've coded a very hacky fix for this. Basically if we're not about to overwrite the file from source, we overwrite it from itself to reset the modification time. A similar effect could be achieved by `touch()`ing the file path but that's outside the concerns of `Cache.php` so I'll let you decide if this is worth committing.

In any case, a decent caching solution seems like it shouldn't be hitting the source every single time just to check modification dates.

---
This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/452)

### DIFF
--- a/src/Gaufrette/Adapter/Cache.php
+++ b/src/Gaufrette/Adapter/Cache.php
@@ -211,8 +211,11 @@ class Cache implements Adapter,
                 if (time() - $this->ttl >= $dateCache) {
                     $dateSource = $this->source->mtime($key);
                     $needsReload = $dateCache < $dateSource;
-                    // touch file mod time so that we only check the source once per $ttl interval
-                    $this->cache->write($key, $this->cache->read($key));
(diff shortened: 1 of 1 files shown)


NiR-:
Sorry @zyphlar for the delay, but we're deprecating Cache adapter (see #464). We'll replace it by a PSR-compliant adapter for 1.0 (#469). Thank you for taking time to create this patch.

PR #457, opened 2017-03-16
Upgrade aws-sdk-php to v3

NiR-:
We're stuck with guzzle/guzzle which is abandoned, because of aws-sdk-php v2.
Hence we need to allow aws-sdk-php v3 which relies on guzzlehttp/guzzle v5/v6.

### DIFF
--- a/tests/Gaufrette/Functional/Adapter/AwsS3Test.php
+++ b/tests/Gaufrette/Functional/Adapter/AwsS3Test.php
@@ -2,41 +2,50 @@
 
 namespace Gaufrette\Functional\Adapter;
 
-use Gaufrette\Adapter\AwsS3;
+use Aws\CommandInterface;
+use Aws\Result;
+use Aws\S3\Exception\S3Exception;
 use Aws\S3\S3Client;
-use Guzzle\Plugin\Mock\MockPlugin;
-use Guzzle\Http\Message\Response;
+use Aws\MockHandler;
+use Gaufrette\Adapter\AwsS3;
 
 /**
  * @todo move to phpspec
  */
 class AwsS3Test extends \PHPUnit_Framework_TestCase
 {
-    protected function getClient()
+    protected function getClient(MockHandler $mock = null)
     {
-        return S3Client::factory(array(
-            'key'    => 'foo',
-            'secret' => 'bar'
-        ));
+        return new S3Client([
+            'handler'     => $mock,
+            'region'      => 'eu-west-1',
+            'version'     => 'latest',
+            'credentials' => [
+                'key'    => 'foo',
+                'secret' => 'bar',
+            ],
+        ]);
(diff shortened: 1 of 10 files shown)


stof:
@NiR- there is no dependency to #459, as this PR allows both 2.x and 3.x of the SDK, meaning that it is still compatible with PHP 5.4 (using the 2.x version in this case)

NiR-:
@stof Yep that's true: nothing has changed in the adapter, meaning either aws-sdk-php v2 or v3 can be used. But in the other hand, the way tests are written has changed with aws-sdk-php v3, so we're not really able to cover both versions (or have you any idea for that issue?).

And I think this PR depends on #459 because this newer version now requires php >= 5.5, nope ?

stof:
why are the functional tests mocking the HTTP client ? they are not functional tests anymore.

NiR-:
@stof Good point, it should not mock HTTP reponses. But [it was already there](https://github.com/KnpLabs/Gaufrette/pull/457/files#diff-61b9bc32613b9bc34d9ea4c28e31f601L25). What would be the best in that case? Run them against a mocked service?

stof:
@NiR- or do like you do for some other functional test: skip them if you don't have credentials provided, and run them against a real bucket otherwise (taking care of cleaning the files created during the testsuite at the end to avoid people paying for dummy files forever, even if small files are cheap)

stof:
and the fact that functional tests are mocking the HTTP client is precisely what makes it hard to support both versions of the SDK in the testsuite, as they use different HTTP client.
Otherwise, it would be easy: detect the available version to change the instantiation of the SDK client, and you're done.

NiR-:
I fixed the functional tests yesterday and I also added env vars for AWS to travis so tests for this adapter will be run on master (or any other branch on this repo, but not for merge).

tests/Gaufrette/Functional/Adapter/AwsS3Test.php
```diff
@@ -2,116 +2,144 @@
...
+        $key = getenv('AWS_KEY');
+        $secret = getenv('AWS_SECRET');
+
+        if (empty($key) || empty($secret)) {
+            $this->markTestSkipped();
+        }
+
+        $this->bucket = uniqid(getenv('AWS_BUCKET'));
+
+        if (self::$SDK_VERSION === 3) {
```
stof: isn't there a way to feature-detect it instead of using the composer metadata ? Maybe with ``method_exists(S3Client::class, 'factory')`` to detect version 2 ?

NiR-: `factory` still exists in version 3, but it is now deprecated. There is a new `S3ClientInterface`, unfortunately it was introduced in version 3.17.0 :/

PR #460, opened 2017-03-20
Replace mongo with mongodb ext

NiR-:


src/Gaufrette/Adapter/GridFS.php
```diff
@@ -18,41 +19,52 @@ class GridFS implements Adapter,
...
-        $id = $this->gridFS->storeBytes($content, $metadata);
-        $file = $this->gridFS->findOne(array('_id' => $id));
+        try {
+            fwrite($stream, $content);
+        } finally {
+            fclose($stream);
+        }
 
-        return $file->getSize();
+        return strlen($content);
```
AntoineLelaisant: We should return the result of `fwrite($stream, $content)` instead because with this implementation there is no way to even know an error occurred.

PR #316, opened 2014-12-12
Invalidate instead of update fileRegister

Sander-Toonen:
After renaming a file, the file in the fileRegister contains the wrong
key. Therefore invalidate the cache instead of updating it.

See #197

---
This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/316)

Issue #197, opened 2013-07-04
[Filesystem] File Object unaware of key change


pawski:
Gaufrette fails on this test (added to Adapter/FunctionalTestCase).

``` php
    public function shouldUpdateFileObjectKeyWhenRenaming()
    {
        $fileObject = $this->filesystem->createFile('somefile');
        $fileObject->setContent('abc'); 

        $this->filesystem->rename('somefile', 'somenewfile');

        $fileObject->setContent('def'); 

        $this->assertFalse($this->filesystem->has('somefile'));
        $this->assertTrue($this->filesystem->has('somenewfile'));
        $this->assertEquals('def', $this->filesystem->read('somenewfile'));
    }
```

Tested on Adapter Local but it is a case of Filesystem.
After renaming operation FileObject key is not touched, so in result we have:
- two files "somefile", "somenewfile"
- content of "somefile" is "def"
- content of "somenewfile" is "abc"

NiR-:
This has been fixed in #468. Thank you for taking to report this issue.

PR #468, opened 2017-03-28
Add rename method to File

NiR-:
It should solve issue describe in #316 in a cleaner way.

---
This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/468)

### DIFF
--- a/src/Gaufrette/File.php
+++ b/src/Gaufrette/File.php
@@ -194,6 +194,18 @@ class File
     }
 
     /**
+     * Rename the file and move it to its new location.
+     *
+     * @param string $newKey
+     */
+    public function rename($newKey)
+    {
+        $this->filesystem->rename($this->key, $newKey);
+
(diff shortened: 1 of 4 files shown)


src/Gaufrette/File.php
```diff
@@ -194,6 +194,17 @@ public function createStream()
     }
 
     /**
+     * @param string $newKey
+     */
+    public function rename($newKey)
+    {
+        // Should be replaced by a try..finally block
```
stof: what should be replaced ?

NiR-:
I'm not totally happy with this PR: there's pretty much nothing to warn about consistency problems in case of direct usage (see #316 & #197). I think the best solution would be to mark `Filesystem::rename` as [`@internal`](https://phpdoc.org/docs/latest/references/phpdoc/tags/internal.html) method. But it's a BC change. What do you think ?

PR #472, opened 2017-04-02
Fixed error message parsing for newest Azure Storage version

nico-incubiq:
Microsoft Azure Storage SDK v0.13.0 has removed the getErrorReason method, which breaks the current error parsing.
I updated the parser, mainly copying what Microsoft used to do, while keeping compatibility with previous versions.

---
This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/472)


src/Gaufrette/Adapter/AzureBlobStorage.php
```diff
@@ -403,4 +408,53 @@ private function guessContentType($content)
...
+     *
+     * @return string
+     */
+    protected static function parseErrorCode(ResponseInterface $response)
+    {
+        //try to parse using xml serializer, if failed, return the whole body
+        //as the error message.
+        try {
+            $sxml = new \SimpleXMLElement($response->getBody());
+            $data = static::_sxml2arr($sxml);
```
NiR-: Since we are only interested in the very first level of `$sxml`, I guess we don't need to recursively convert the object.

src/Gaufrette/Adapter/AzureBlobStorage.php
```diff
@@ -403,4 +408,53 @@ private function guessContentType($content)
...
+     * @param  ResponseInterface $response The response with a response body.
+     *
+     * @return string
+     */
+    protected static function parseErrorCode(ResponseInterface $response)
+    {
+        //try to parse using xml serializer, if failed, return the whole body
+        //as the error message.
+        try {
+            $sxml = new \SimpleXMLElement($response->getBody());
```
NiR-: `$sxml` is a weird name. `$xml` could do the job, but `$body` would be better.

src/Gaufrette/Adapter/AzureBlobStorage.php
```diff
@@ -403,4 +408,53 @@ private function guessContentType($content)
...
+            $data = static::_sxml2arr($sxml);
+
+            if (array_key_exists('Code', $data)) {
+                $errorCode = $data['Code'];
+            } else {
+                $errorCode = $response->getReasonPhrase();
```
NiR-: You don't need this `else` clause as soon as you define `$response->getResponsePhrase()` as a default value for `$errorCode`, at the top of this method. By doing this, you can also delete the assignment in the `catch`.

src/Gaufrette/Adapter/AzureBlobStorage.php
niko3oo: Done

src/Gaufrette/Adapter/AzureBlobStorage.php
niko3oo: Done

src/Gaufrette/Adapter/AzureBlobStorage.php
niko3oo: I would prefer to leave it as is, just in case we need more information later.
Anyway I mainly copied what Azure storage bundle does on its side.

NiR-:
I'm still not totally happy with your PR: I don't really like copy/paste especially because I see adding dead code as a bad practice. Thus I think it would be more beneficial to add a `@see` link plus further details about why it's needed. For instance, a little note about which version of the SDK introducing this BC change would be really great.

In any case, thank you for reporting and fixing this issue :) Would be glad to see it merged before the next release (coming next week)!

nico-incubiq:
Sorry for the delay, I do not have access to the project I used to work on, so I cannot test / improve my fix anymore. Feel free to add comments to it if needed. Last time I checked it worked well with Azure Storage 0.13.0. Btw I just fixed the merge issues.

NiR-:
@nico-incubiq No problem, I can run Azure tests on my side. Since #486 was merged it can happens that Azure SDK throws exceptions we ignore. Hence I think we should really avoid parsing unneeded xml fields. Could you fix this please? I can do it myself if you prefer (still you will be credited for the original fix).

nico-incubiq:
Hi @NiR- I fixed the code according to your concerns. I am not sure it runs though, as I do not have the necessary setup to test. I would be glad if you could try to run it on your side and come back to me if you find anything buggy.

NiR-:
@nico-incubiq I tested your patch and unfortunately it is broken :/ After digging into azure code I realise it conflicts with #486, because the following issues had been addressed in that previous PR. Really sorry for wasting your time, I didn't realise this until now :/

* `ServicesException::getErrorReason()` has been removed since microsoft/azure-storage v0.13.0
* `ServicesException` has been moved to `MicrosoftAzure\Storage\Common\Exceptions\ServiceException` since microsoft/azure-storage v0.14.0
* `CreateBlobOptions::setBlobContentType` has been renamed into `setContentType` since v0.15.0.

 I'll pinpoint microsoft/azure-storage to v0.15.0 in [the new metapackage](https://github.com/Gaufrette/azure-blob-storage-adapter).

PR #477, opened 2017-04-11
Don't use FTP_USEPASSVADDRESS before php 5.6.18

NiR-:
See https://github.com/php/php-src/commit/90a26a4844d1acfecb3bd0842da333f8bddd45c6

---
This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/477)


stof:
this is incomplete. For PHP 7, this commit was part of 7.0.2 only, not of 7.0.0

stof:
other solution: use feature detection by using ``if (defined('FTP_USEPASVADDRESS'))`` rather than doing version sniffing

NiR-:
Ha indeed I missed that one. I'll fix it in another PR. Thank you @stof.

PR #479, opened 2017-04-13
Replace file_exists with is_file. Fixes #473

### LINKED ISSUE #473: Local adapter read doesn't throw an error when accidentally reading a directory
We've had the case that file was stored properly but only a path like `/some/path/' was stored in the DB. The file was actually saved to the local file system.

So what happens now when passing that path to read() is that you get an empty result back but no exception like when the file is not found.

I've checked the code https://github.com/KnpLabs/Gaufrette/blob/master/src/Gaufrette/Adapter/Local.php#L80 and it is using file_exists() which doesn't make a difference between file and folder, so it was reporting true back.

The method is missing an is_dir() check.

AntoineLelaisant:
Fixes #473

---
This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/479)

### DIFF
--- a/spec/Gaufrette/Adapter/LocalSpec.php
+++ b/spec/Gaufrette/Adapter/LocalSpec.php
@@ -156,6 +156,6 @@ class LocalSpec extends ObjectBehavior
     {
         $this->beConstructedWith(vfsStream::url('test/other'), true);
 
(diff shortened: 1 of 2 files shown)


NiR-:
@AntoineLelaisant Your patch is broken ;)

Issue #465, opened 2017-03-24
Run as much functional tests as possible on Travis


NiR-:
We've currently many functional tests ignored on Travis because of missing backend services (for instance tests for FTP and AWS are not run on Travis). We need to improve this situation in order to prevent merging broken patches.

What's needed:
* Host, port, credentials, ... should be configurable via phpunit.xml and environment variables
* Tests should be skipped with an explicit message if required env vars are not defined
* Install required services on Travis, and update .travis.yml accordingly
* If necessary/possible add a docker container matching Travis environment, so what runs on Travis can be run locally

The best would be to run tests for every maintained adapter (see #464), meaning:
* [x] AwsS3
* [x] AzureBlobStorage (see #500)
* [x] DoctrineDbal (see #504)
* ~Flysystem (already has specs, functional tests won't be valuable)~
* [x] Ftp (see #483)
* [x] GridFS
* [x] OpenCloud (see #505)
* [x] PhpseclibSftp (see #484)

PR #484, opened 2017-04-26
Run functional tests for PhpseclibSftp adapter on travis

NiR-:
Related to #465.

This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/484)


tests/Gaufrette/Functional/Adapter/PhpseclibSftpTest.php
```diff
@@ -3,7 +3,44 @@
...
+
+    public function setUp()
+    {
+        $host     = getenv('SFTP_HOST');
+        $port     = getenv('SFTP_PORT') ?: 22;
+        $user     = getenv('SFTP_USER');
+        $password = getenv('SFTP_PASSWORD');
+        $baseDir  = getenv('SFTP_BASE_DIR');
+
+        if ($host === false || $user === false || $password === false || $baseDir === false) {
```
NiR-: That's a long debate :p But generally speaking I'm against it: (strict) equality is the only comparison operator for which yoda conditions might make sense. Thus it introduces more oddities than anything else. And in the meantime, I think what could be solved by yoda conditions should be solved by code review.

Issue #446, opened 2017-02-22
[Feature] Support for multi-container Azure workflow


dkarlovi:
Currently, the `AzureBlobStorage` adapter requires a single container name which will be filesystem root, it might try to create it.

If I'm using a single storage account to hold only one type of asset (say, only images), I'd like to use one container per image (which will hold the original and all the variants), much like for example Azure Media Services does. This also [seems to be a recommended way of doing things](http://stackoverflow.com/a/8159446/672885).

What could be done here is add an additional flag / mode which would take the storage key and use the part until the first slash, create a container (if it doesn't already exist) and use the rest of the path as key. Ie. 

    write('my/image/path.jpeg') // "my" = container, "image/path.jpeg" = key

Some sort of container default creation options should also be provided (for example, access mode), but I think it shouldn't be all that hard.

NiR-:
Hello @dkarlovi,

[The `AzureBlobStorage` adapter already has a `$containerName` argument, and a `$create` one.](https://github.com/KnpLabs/Gaufrette/blob/master/src/Gaufrette/Adapter/AzureBlobStorage.php#L55).

Is this what you are looking for? Or is it something else?

dkarlovi:
Hi @NiR-,

that's not really what I'm looking for. Having a container name predefined means you can only have one container at a time, where I'd like to have dynamic containers where each asset has its own container for its use (like, for example, Azure Media Services does).

Also, having `create: true` is somewhat clumsy as it's hard-coded to config: if you leave it to `true`, it throws an exception on each request except the first (at least it did in the last version I've tried). If you leave it to `false`, you can't just switch your Azure Blob Storage account, you need to do some maintenance upfront before you can start using it (create the container in question).

My proposed feature here would allow for containers behaving as "first level directories", where they would get auto-created if not already there, with adapter specific container options, for example having it be public by default.

NiR-:
@dkarlovi ok, now I see what you're looking for: basically, for every write operation we would have to check if the container (first directory in the path) exists, right ?

In your opinion what would be the impact on billing and performance of such a strategy? If it's negligable for both the best would be to update the current adapter. In other case, I guess it would be better to create  a seperated adapter.

dkarlovi:
@NiR- yes, exactly like that.

I don't think it would impact billing as (AFAIK) storage is billed per consumed asset size and transfer bandwidth, not counting API operations or metadata (containers are just metadata, they don't count as data). This switch should have no impact what so ever.

[...]

PR #486, opened 2017-04-29
Add support for Azure multicontainer

dkarlovi:
#446

---
This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/486)

### DIFF
--- a/src/Gaufrette/Adapter/AzureBlobStorage.php
+++ b/src/Gaufrette/Adapter/AzureBlobStorage.php
@@ -5,11 +5,13 @@ namespace Gaufrette\Adapter;
 use Gaufrette\Adapter;
 use Gaufrette\Util;
 use Gaufrette\Adapter\AzureBlobStorage\BlobProxyFactoryInterface;
+use MicrosoftAzure\Storage\Blob\Models\Blob;
+use MicrosoftAzure\Storage\Blob\Models\Container;
 use MicrosoftAzure\Storage\Blob\Models\CreateBlobOptions;
 use MicrosoftAzure\Storage\Blob\Models\CreateContainerOptions;
 use MicrosoftAzure\Storage\Blob\Models\DeleteContainerOptions;
 use MicrosoftAzure\Storage\Blob\Models\ListBlobsOptions;
-use MicrosoftAzure\Storage\Common\ServiceException;
+use MicrosoftAzure\Storage\Common\Exceptions\ServiceException;
 
 /**
  * Microsoft Azure Blob Storage adapter.
@@ -47,19 +49,49 @@ class AzureBlobStorage implements Adapter,
     protected $blobProxy;
 
     /**
+     * @var bool
+     */
+    protected $multiContainerMode = false;
+
+    /**
+     * @var CreateContainerOptions
+     */
+    protected $createContainerOptions;
+
+    /**
      * @param AzureBlobStorage\BlobProxyFactoryInterface $blobProxyFactory
-     * @param string                                     $containerName
+     * @param string|null                                $containerName
      * @param bool                                       $create
      * @param bool                                       $detectContentType
+     *
+     * @throws \RuntimeException
      */
-    public function __construct(BlobProxyFactoryInterface $blobProxyFactory, $containerName, $create = false, $detectContentType = true)
+    public function __construct(BlobProxyFactoryInterface $blobProxyFactory, $containerName = null, $create = false, $detectContentType = true)
     {
         $this->blobProxyFactory = $blobProxyFactory;
         $this->containerName = $containerName;
         $this->detectContentType = $detectContentType;
-        if ($create) {
+        if (null === $containerName) {
+            $this->multiContainerMode = true;
+        } elseif ($create) {
             $this->createContainer($containerName);
         }
+    }
+
+    /**
+     * @return CreateContainerOptions
+     */
+    public function getCreateContainerOptions()
+    {
+        return $this->createContainerOptions;
+    }
+
+    /**
+     * @param CreateContainerOptions $options
+     */
+    public function setCreateContainerOptions(CreateContainerOptions $options)
+    {
+        $this->createContainerOptions = $options;
     }
 
     /**
@@ -74,12 +106,16 @@ class AzureBlobStorage implements Adapter,
     {
         $this->init();
 
+        if (null === $options) {
+            $options = $this->getCreateContainerOptions();
+        }
+
         try {
             $this->blobProxy->createContainer($containerName, $options);
         } catch (ServiceException $e) {
             $errorCode = $this->getErrorCodeFromServiceException($e);
 
-            if ($errorCode != self::ERROR_CONTAINER_ALREADY_EXISTS) {
+            if ($errorCode !== self::ERROR_CONTAINER_ALREADY_EXISTS) {
                 throw new \RuntimeException(sprintf(
                     'Failed to create the configured container "%s": %s (%s).',
                     $containerName,
@@ -107,7 +143,7 @@ class AzureBlobStorage implements Adapter,
         } catch (ServiceException $e) {
             $errorCode = $this->getErrorCodeFromServiceException($e);
 
-            if ($errorCode != self::ERROR_CONTAINER_NOT_FOUND) {
+            if ($errorCode !== self::ERROR_CONTAINER_NOT_FOUND) {
                 throw new \RuntimeException(sprintf(
                     'Failed to delete the configured container "%s": %s (%s).',
                     $containerName,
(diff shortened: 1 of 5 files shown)


dkarlovi:
@NiR- do you think this direction is OK?

dkarlovi:
@NiR- bump.

NiR-:
Sorry, I was in vacation last week. I'll check it today ;)

dkarlovi:
Sorry to bump, it's not urgent or anything, just don't want it to get lost. Hope you had a great vacation. :+1:

tests/Gaufrette/Functional/adapters/AzureMultiContainerBlobStorage.php.dist
```diff
@@ -0,0 +1,6 @@
+<?php
```
NiR-: We're deprecating this way of running functional tests. Could you rather implement `setUp` / `tearDow` methods to initialize the filesystem in your Test please? [Here's an example](https://github.com/KnpLabs/Gaufrette/blob/master/tests/Gaufrette/Functional/Adapter/AwsS3Test.php#L23).

src/Gaufrette/Adapter/AzureBlobStorage.php
```diff
@@ -403,4 +481,26 @@ private function guessContentType($content)
...
+    /**
+     * @param string $key
+     *
+     * @return array
+     * @throws \InvalidArgumentException
+     */
+    private function tokenizeKey($key)
+    {
+        $containerName = $this->containerName;
+        if (true === $this->multiContainerMode) {
```
NiR-: I would rather invert the conditional ` true === $this->multiContainerMode` to early return a value when it's disabled. It would make it a bit more readable.

tests/Gaufrette/Functional/Adapter/AzureMultiContainerBlobStorageTest.php
```diff
@@ -0,0 +1,178 @@
+<?php
+
+namespace Gaufrette\Functional\Adapter;
+
+/**
+ * Class AzureMultiContainerBlobStorageTest
+ * @group AzureBlobStorage
+ * @group AzureMultiContainerBlobStorage
+ */
+class AzureMultiContainerBlobStorageTest extends FunctionalTestCase
```
NiR-: As you're overriding test cases, functional test for `multiContainerMode` disabled are not run. I think it would be better to instantiate two filesystems here and create new test cases for multi container mode.

dkarlovi:
@NiR- thanks! :+1:

tests/Gaufrette/Functional/Adapter/AzureMultiContainerBlobStorageTest.php
dkarlovi: I don't understand this one. I've added a new test suite, haven't disabled the existing one. By default, both will be run (and I have run both test sets).

dkarlovi: $ vendor/bin/phpunit --group AzureBlobStorage --testdox
    PHPUnit 3.7.38 by Sebastian Bergmann.
    
    Configuration read from /home/dado/Development/Gaufrette/phpunit.xml.dist
    
    Gaufrette\Functional\Adapter\AzureBlobStorage
    [x] Should write and read
    [x] Should update file content
    [x] Should check if file exists
    [x] Should get mtime
    [x] Should fail when try mtime for key which does not exist
    [x] Should rename file
    [x] Should delete file
    [x] Should fetch keys
    [x] Should work with hidden files
    [x] Should keep file object in register
    [x] Should wrtie to same file
    
    Gaufrette\Functional\Adapter\AzureMultiContainerBlobStorage
    [x] Should write and read
    [x] Should update file content
    [x] Should check if file exists
    [x] Should get mtime
    [x] Should fail when try mtime for key which does not exist
    [x] Should rename file
    [x] Should delete file
    [x] Should fetch keys
    [x] Should work with hidden files
    [x] Should keep file object in register
    [x] Should wrtie to same file

tests/Gaufrette/Functional/adapters/AzureMultiContainerBlobStorage.php.dist
dkarlovi: Done.

tests/Gaufrette/Functional/Adapter/AzureMultiContainerBlobStorageTest.php
dkarlovi: Everything still passes with the last commit.

NiR-: Haa indeed, sorry :+1:

tests/Gaufrette/Functional/adapters/AzureMultiContainerBlobStorage.php.dist
NiR-: Thank you :) I will set up Azure Storage to ensure tests against code landing on our repository tests are are run.

NiR-:
@dkarlovi :+1: Could you squash "wip" commits, and in the meantime, add a note in the `doc/` about this new feature please? Would be glad to see it merged :)

dkarlovi:
Will do, thanks.

BTW I've noticed one problem still, will need to fix it too. Azure SDK guys keep changing the interface, for example they've renamed `setBlobContentType()` to `setContentType()` (which is used if `detectContentType=true`).

Adapters should really get their own composer.json so they can lock the SDK dependency down. :(

NiR-:
@dkarlovi We'll release a new version of this repo. From that release we'll recommend developers to require newly created metapackages rather than this repo. So actually we can lock Azure SDK version out there :)

dkarlovi:
@NiR- So you've decided against the mono-repo?

NiR-:
@dkarlovi Currently we're in favor of mono-repo because we think it will be simpler for us to overhaul the project. We might split the repo later though. So yep you're right, it's not ideal because information about supported version is spreaded across many repositories. In the mean we consider it as a little drawback. We can overcome this issue by adding a note in the README + locking the version in `require-dev` section.

Which version introduce the BC break you're talking about? Btw, thank you for reporting it :)

dkarlovi:
@NiR- done, I've squashed and rebased to latest master.

src/Gaufrette/Adapter/AzureBlobStorage.php
```diff
@@ -47,22 +49,52 @@ class AzureBlobStorage implements Adapter,
...
+     */
+    public function getCreateContainerOptions()
+    {
+        return $this->createContainerOptions;
+    }
+
+    /**
+     * @param CreateContainerOptions $options
+     */
+    public function setCreateContainerOptions(CreateContainerOptions $options)
```
NiR-: IMHO it would be better to have this as constructor parameter, it might even replace current `$create` argument: it could either take a `CreateContainerOptions` instance, meaning container(s) should be created if needed, or null, meaning it should fail if container(s) does not exist. But it would be a BC change (actually it would be better in another PR). What do you think?

dkarlovi: I agree actually, did this to keep BC. 👍 

If null, it will be treated as an empty `CreateContainerOptions`, meaning containers get created with whatever the default values other than name are. We can go with this for now and revamp the whole adapter interface with 1.0?

src/Gaufrette/Adapter/AzureBlobStorage.php
```diff
@@ -120,78 +156,91 @@ public function deleteContainer($containerName, DeleteContainerOptions $options
...
 
-            $this->blobProxy->createBlockBlob($this->containerName, $key, $content, $options);
+            $options->setContentType($contentType);
+        }
 
-            if (is_resource($content)) {
-                return Util\Size::fromResource($content);
-            }
+        try {
+            $this->createContainer($containerName);
```
NiR-: Doing a final review I realized this should be done only when `multiContainerMode` is enabled. Also I think `$this->create` should be checked too, so there's no big differences introduced between mono/multi container mode. Beside this point, great job!

src/Gaufrette/Adapter/AzureBlobStorage.php
NiR-: Yep, that's what I thought :)

src/Gaufrette/Adapter/AzureBlobStorage.php
dkarlovi: Multi-container mode ignores `$create` as it wouldn't make much sense to say "Create containers automatically, but don't". :) Agree to check for multi-mode here, will do it today. 👍 

Thanks, it's for my selfish interest to have this feature, obviously, but glad there's a Gaufrette to contribute it to in the first place, thank you all for that.

dkarlovi: @NiR- done, also for rename()

PR #488, opened 2017-05-18
Local stream does not use the mode defined by Local adapter

NiR-:
Local stream does not use the mode defined by Local adapter and enforces 0755.  For more background see #320.

Credits goes to @richardfullmer, I just added tests as requested in the original PR.

---
This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/488)

### DIFF
--- a/tests/Gaufrette/Functional/LocalDirectoryDeletor.php
+++ b/tests/Gaufrette/Functional/LocalDirectoryDeletor.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace Gaufrette\Functional;
+
+class LocalDirectoryDeletor
+{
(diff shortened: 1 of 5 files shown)


tests/Gaufrette/Functional/LocalDirectoryDeletor.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace Gaufrette\Functional;
+
+class LocalDirectoryDeletor
+{
+    public static function deleteDirectory($directory)
+    {
```
fabschurt: Would it make sense to check that `$directory` is not equal to `/`, as a basic fail-safe?

PR #503, opened 2017-06-08
Remove support for php 5.4 & php 5.5

NiR-:
This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/503)


composer.json
```diff
@@ -23,7 +23,8 @@
         }
     ],
     "require": {
-        "php": ">=5.4"
+        "php": ">=5.6",
+        "akeneo/phpspec-skip-example-extension": "~1.2"
```
Shivoham: It should be a dev requirement

tests/Gaufrette/Functional/Adapter/AzureMultiContainerBlobStorageTest.php
```diff
@@ -182,14 +200,28 @@ public function shouldKeepFileObjectInRegister()
...
         $FileObjectA->setContent('ABC');
 
-        $FileObjectB = $this->filesystem->createFile('container11/somefile');
+        $FileObjectB = $this->filesystem->createFile($path);
         $FileObjectB->setContent('DEF');
 
         $this->assertEquals('DEF', $FileObjectB->getContent());
+    }
+
+    private function getContainerName($prefix)
```
Shivoham: It's not obivous this method won't return the same result when called with the same args. I'd prefer something like `createUniqueContainerName($prefix)`

NiR-:
@dkarlovi From time to time `AzureMultiContainerBlobStorageTest::shouldWorkWithHiddenFiles()` fails with following error:
> RuntimeException: Failed to retrieve keys: container "" not found.

Have you any idea why this happen?

dkarlovi:
I think we should add a test which fetches keys from a completely empty storage (no containers, no blobs), that might be the issue here. Would need to see the actual exception getting caught there though.

PR #505, opened 2017-06-14
Opencloud tests

NiR-:
This change is [Reviewable](https://reviewable.io/reviews/knplabs/gaufrette/505)


NiR-:
@dkarlovi I disabled AzureMultiContainerBlobStorageTest since it was too flaky. I'll update how exceptions are thrown in the lib, thus I think it would be easier to debug after that. I'll re-enable it only when it's fixed.

dkarlovi:
Sounds good, mention me to help fix it once you're done.

PR #512, opened 2017-06-26
Azure tests on appveyor

NiR-:
I forgot about this stale branch, but I wanted to finish it before releasing v0.4, too bad :/ I create a WIP pull request to remind me about it.

Currently appveyor builds are useless because they're always green. In fact tests are not executed on it: looking at [this build](https://ci.appveyor.com/project/NiR-/gaufrette/build/1.0.198#L1051) for instance, we can see that phpspec, phpunit and the end of build happened at the exact same second, `07:50` (appears at hover).

This PR tries another way of installing php and required extension. IIRC, I couldn't install mongodb extension with the previous setup. Both problem should be addressed but I'm not totally sure since I started my investigations almost a month ago :(

### DIFF
--- a/tests/Gaufrette/Functional/Adapter/FunctionalTestCase.php
+++ b/tests/Gaufrette/Functional/Adapter/FunctionalTestCase.php
@@ -67,8 +67,6 @@ EOF
 
         $this->assertEquals('Some content', $this->filesystem->read('foo'));
         $this->assertEquals('Some content1', $this->filesystem->read('test/subdir/foo'));
-        $this->filesystem->delete('foo');
-        $this->filesystem->delete('test/subdir/foo');
     }
 
     /**
@@ -81,7 +79,6 @@ EOF
         $this->filesystem->write('foo', 'Some content updated', true);
 
         $this->assertEquals('Some content updated', $this->filesystem->read('foo'));
-        $this->filesystem->delete('foo');
     }
 
     /**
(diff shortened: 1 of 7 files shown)


NiR-:
The last build on appveyor is green, and tests are really executed this time.

tests/Gaufrette/Functional/Adapter/FunctionalTestCase.php
```diff
@@ -67,8 +67,6 @@ public function shouldWriteAndRead()
 
         $this->assertEquals('Some content', $this->filesystem->read('foo'));
         $this->assertEquals('Some content1', $this->filesystem->read('test/subdir/foo'));
-        $this->filesystem->delete('foo');
```
nicolasmure: Are the removal to the ` $this->filesystem->delete` calls mandatory ? (because it also affects the other adapter tests)

NiR-: Yep there are useful because it's not the place where such clean up tasks should happen: it makes tests flaky since any failure would mark test as failed, even if it's not. I've seen such behavior with Azure adapter. Also every maintained adapter has a dedicated cleanup procedure, and IMHO other tests should be removed.

NiR-:
I rebased it. Could you review it, or is it good for you, @nicolasmure?

NiR-:
Thank you :)

Issue #497, opened 2017-06-02
Use exceptions if error(s) happen(s)


NiR-:
Currently we return false (or something similar) to notify error(s). It leads to bad DX. We should improve this for v1.0.

NiR-:
We need to remove support of php v5.6 if we want to properly handle errors of Local adapter. Also OpenCloud adapter shouldn't be updated for now: it's based on an outdated SDK.

PR #517, opened 2017-07-07
Introduce StorageFailure and FileNotFound exceptions

NiR-:
The main reason: it's currently impossible to know what's going wrong when one of the method fails. So rather than bad return values, I updated the Adapter API to throw exceptions.

So, this PR introduces two exceptions to be used by adapters: `Gaufrette\Exception\StorageFailure` and `Gaufrette\Exception\FileNotFound`. I implemented only AwsS3 for now and Filesystem methods would benefit from such overhauling (something like `FilesystemException`). But this patch has been done mostly to demonstrates what we could do to enhance the current API, more will come if this one is accepted.

See #497.

### DIFF
--- a/src/Gaufrette/Exception/StorageFailure.php
+++ b/src/Gaufrette/Exception/StorageFailure.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace Gaufrette\Exception;
+
+use Gaufrette\Exception;
+
+/**
+ * Exception thrown when an unexpected error happened at the storage level (or its underlying sdk).
+ *
+ * @author Albin Kerouanton <albin.kerouanton@knplabs.com>
+ */
+class StorageFailure extends \RuntimeException implements Exception
+{
+    /**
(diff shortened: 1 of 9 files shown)


src/Gaufrette/Exception/StorageFailure.php
```diff
@@ -0,0 +1,28 @@
...
+use Gaufrette\Exception;
+
+/**
+ * Exception thrown when an unexpected error happened at the storage level (or its underlying sdk).
+ *
+ * @author Albin Kerouanton <albin.kerouanton@knplabs.com>
+ */
+class StorageFailure extends \RuntimeException implements Exception
+{
+    public static function unexpectedFailure($action, $args, \Exception $previous = null)
```
nicolasmure: Can you type hint the `$args` parameter (and also add a doc block) please ?

NiR-:
I added the missing phpdoc and typehint ;)

nicolasmure:
Thank you ! :+1:

PR #519, opened 2017-07-10
StorageFailure and FileNotFound exceptions for Flysystem adapter

nicolasmure:
Related to #497 .

To be complient with the new [Adapter implementation](https://github.com/KnpLabs/Gaufrette/pull/517), the adapter throws a `StorageFailure` exeption instead of returning `false` if the original adapter returns `false` in case of failure.

### DIFF
--- a/spec/Gaufrette/Adapter/FlysystemSpec.php
+++ b/spec/Gaufrette/Adapter/FlysystemSpec.php
@@ -3,9 +3,9 @@
 namespace spec\Gaufrette\Adapter;
 
 use PhpSpec\ObjectBehavior;
-use Prophecy\Argument;
 use League\Flysystem\AdapterInterface;
 use League\Flysystem\Config;
+use League\Flysystem\FileNotFoundException;
 
 class FlysystemSpec extends ObjectBehavior
 {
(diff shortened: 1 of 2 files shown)


spec/Gaufrette/Adapter/FlysystemSpec.php
```diff
@@ -31,11 +31,46 @@ function it_reads_file(AdapterInterface $adapter)
         $this->read('filename')->shouldReturn('Hello.');
     }
 
+    function it_throws_file_not_found_exception_when_trying_to_read_an_unexisting_file(AdapterInterface $adapter)
+    {
+        $adapter->read('filename')->willThrow(new FileNotFoundException('filename'));
+
+        $this->shouldThrow('Gaufrette\Exception\FileNotFound')->duringread('filename');
```
nicolasmure: It looks surprising but yes, it's because of an old version of phpspec (2.4). BTW I ran into issues when trying to run the phpspec tests : the [`phpspec.yml`](https://github.com/KnpLabs/Gaufrette/blob/master/phpspec.yml) file has a dependency on [`Akeneo\SkipExampleExtension`](https://github.com/akeneo/PhpSpecSkipExampleExtension/blob/master/src/Akeneo/SkipExampleExtension.php) but I couldn't get it installed via composer (unresolvable set of packages). So I had to comment that file to run the tests on my machine :confused: . I'm still wondering how the tests can be executed on the CI .

It would be a good point to bump up the tests tools to their latest version :wink:

NiR-: > I ran into issues when trying to run the phpspec tests : the phpspec.yml file has a dependency on Akeneo\SkipExampleExtension but I couldn't get it installed via composer (unresolvable set of packages). So I had to comment that file to run the tests on my machine :confused:

Mh, that's strange! The `akeneo/phpspec-skip-example-extension` package [is in our composer.json](akeneo/phpspec-skip-example-extension) for some time now. Was there any error message/stacktrace?

> It would be a good point to bump up the tests tools to their latest version :wink:

Yep, I already upgraded phpunit to 5.7 and there's a task about upgrading phpspec in the github project ;)

NiR-:
Great job! Thank you :)

spec/Gaufrette/Adapter/FlysystemSpec.php
nicolasmure: There was no error during `composer install`, but the akeneo directory was not present in the `vendor`. However, when trying to install manually the package, I gor the following error : 
```
$ composer require --dev akeneo/phpspec-skip-example-extension
Using version ^3.0@dev for akeneo/phpspec-skip-example-extension
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - akeneo/phpspec-skip-example-extension 3.0.x-dev requires phpspec/phpspec ^4.0 -> satisfiable by phpspec/phpspec[4.0.x-dev].
    - akeneo/phpspec-skip-example-extension v3.0.0-alpha1 requires phpspec/phpspec ^4.0 -> satisfiable by phpspec/phpspec[4.0.x-dev].
    - Conclusion: don't install phpspec/phpspec 4.0.x-dev
    - Installation request for akeneo/phpspec-skip-example-extension ^3.0@dev -> satisfiable by akeneo/phpspec-skip-example-extension[3.0.x-dev, v3.0.0-alpha1].


Installation failed, reverting ./composer.json to its original content.
```

I haven't noticed the phpspec upgrade card, I'll try to give it a look.

NiR-: Have you try to delete your local `composer.lock` first?

nicolasmure: Shame on me :hammer: removing the `composer.lock` file and running `composer install` again did the trick :) Thank you :+1:

PR #520, opened 2017-07-11
update phpspec to "^3.0"

nicolasmure:
Related to https://github.com/KnpLabs/Gaufrette/projects/1#card-3672083 .


NiR-:
IIRC phpdocs were the way to inject stubs to spec examples with phpspec2. I think we can remove them.

PR #521, opened 2017-08-08
[1.0] Remove meaningless return values & throw exceptions when needed

NiR-:
So far, write, delete and rename methods returned values: number of written bytes
for write, and boolean value for the two other methods. This is no longer the case.

The Local adapter was the first adapter to be introduced (see eed10e3),
and presumably the Adapter interface has been based on it. Hence others adapters
try to mimic low-level file functions. In the end it looks like a bad decision:
most adapters are for object storages, not filesystems. Moreover it forces most
adapters to recompute the size of the written content locally.

Considering CQS principle, it's also better to not return any value from
writting methods.

Related to #497.


NiR-:
Nope, in the end I think it's better to keep `FileNotFound` exception. I re-added it to adapters, excludng: OpenCloud, Ftp and PhpspeclibSftp.

src/Gaufrette/Adapter/Local.php
```diff
@@ -182,51 +168,62 @@ public function createStream($key)
...
     {
+        if (!$this->exists($key)) {
+            throw new FileNotFound($key);
+        }
+
         $fileInfo = new \finfo(FILEINFO_MIME_TYPE);
 
-        return $fileInfo->file($this->computePath($key));
+        if (false === $mimeType = $fileInfo->file($this->computePath($key))) {
+            throw StorageFailure::unexpectedFailure('size', ['key' => $key]);
```
nicolasmure: `size` => `mimeType`

src/Gaufrette/FilesystemInterface.php
```diff
@@ -12,6 +12,7 @@
      * @return bool TRUE if the file exists, FALSE otherwise
      *
      * @throws \InvalidArgumentException If $key is invalid
+     * @throws Exception\StorageFailure  When the underlying storage fails asserting file exists
```
nicolasmure: `fails to assert`

src/Gaufrette/FilesystemInterface.php
```diff
@@ -27,7 +28,7 @@ public function has($key);
      *
      * @throws Exception\FileNotFound    when sourceKey does not exist
      * @throws Exception\UnexpectedFile  when targetKey exists
-     * @throws \RuntimeException         when cannot rename
+     * @throws Exception\StorageFailure  When the underlying storage fails renaming the file
```
nicolasmure: `fails to rename`

and so on

PR #523, opened 2017-08-15
Added support for calculated size for Azure Blob Storage

gerkestam2:
Added the SizeCalculator to the Azure Blob Storage. Size will be retrieved from the blob properties, instead of loading the complete file into memory in order to determine the filesize. 

Tested with a local environment.

### DIFF
--- a/src/Gaufrette/Adapter/AzureBlobStorage.php
+++ b/src/Gaufrette/Adapter/AzureBlobStorage.php
@@ -20,7 +20,9 @@ use MicrosoftAzure\Storage\Common\Exceptions\ServiceException;
  * @author Paweł Czyżewski <pawel.czyzewski@enginewerk.com>
  */
 class AzureBlobStorage implements Adapter,
-                                  MetadataSupporter
+                                  MetadataSupporter,
+                                  SizeCalculator
+
 {
     /**
      * Error constants.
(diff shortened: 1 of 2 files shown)


src/Gaufrette/Adapter/AzureBlobStorage.php
```diff
@@ -308,6 +310,26 @@ public function mtime($key)
...
+        list($containerName, $key) = $this->tokenizeKey($key);
+
+        try {
+            $properties = $this->blobProxy->getBlobProperties($containerName, $key);
+
+            return $properties->getProperties()->getContentLength();
+        } catch (ServiceException $e) {
+            $this->failIfContainerNotFound($e, sprintf('read content length for key "%s"', $key), $containerName);
+
+            return false;
```
nicolasmure: Be careful, neither the [`SizeCalculator`](https://github.com/KnpLabs/Gaufrette/blob/master/src/Gaufrette/Adapter/SizeCalculator.php#L17) nor the [`FilesystemInterface`](https://github.com/KnpLabs/Gaufrette/blob/master/src/Gaufrette/FilesystemInterface.php#L138) have a `bool` return value. Maybe you throw a `\RuntimeException` here instead of returning false ?

NiR-: Actually, only `FTP` and `AwsS3` implements this interface and both differs: the former throws an exception whereas the latter returns false... So I think we can consider this acceptable and it'll be replaced by a proper exception when backported to 1.0 branch.

nicolasmure: :+1: for proper impl in 1.0

tests/Gaufrette/Functional/Adapter/AzureMultiContainerBlobStorageTest.php
```diff
@@ -98,6 +98,19 @@ public function shouldGetMtime()
...
      * @test
      * @group functional
+     */
+    public function shouldGetSize()
+    {
+        $path = $this->createUniqueContainerName('container') . '/foo';
+
+        $this->filesystem->write($path, 'Some content');
+
+        $this->assertGreaterThan(11, $this->filesystem->size($path));
```
NiR-: You should assert about the exact size of the file.

gerkestam2: Thanks, updated the testcode.

nicolasmure:
Thank you @gerkestam2 !

PR #526, opened 2017-09-12
Fix directory deletion

### LINKED ISSUE #524: Local adapter cant delete empty directory
While using Local adapter when i try to delete empty directory i get issue when check happens if key exists. The check for key only checks if it is file - https://github.com/KnpLabs/Gaufrette/blob/master/src/Gaufrette/Adapter/Local.php#L94

I think this is not proper since directories should also be considered a keys and should be deleted if they are empty. Is this a bug or intended like that?

Lctrs:
Close #524 #473.

### DIFF
--- a/src/Gaufrette/Adapter/Local.php
+++ b/src/Gaufrette/Adapter/Local.php
@@ -53,6 +53,10 @@ class Local implements Adapter,
      */
     public function read($key)
     {
+        if ($this->isDirectory($key)) {
+            return false;
+        }
+
         return file_get_contents($this->computePath($key));
     }
 
@@ -91,7 +95,7 @@ class Local implements Adapter,
      */
     public function exists($key)
     {
-        return is_file($this->computePath($key));
+        return file_exists($this->computePath($key));
     }
 
     /**
@@ -147,11 +151,13 @@ class Local implements Adapter,
      */
     public function delete($key)
     {
-        if ($this->isDirectory($key)) {
-            return rmdir($this->computePath($key));
+        $file = $this->computePath($key);
+
+        if (!file_exists($file)) {
+            return false;
         }
 
-        return unlink($this->computePath($key));
+        return $this->doDelete($file);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


Lctrs:
@nicolasmure @NiR- Currently, the fix changes the current behaviour from returning false when trying to delete a non empty directory to deleting the directory recursively.

Do you accept this or do you want to keep the old behaviour ?

Lctrs:
Rebased and tests are green.

src/Gaufrette/Adapter/Local.php
```diff
@@ -91,7 +95,7 @@ public function rename($sourceKey, $targetKey)
      */
     public function exists($key)
     {
-        return is_file($this->computePath($key));
+        return file_exists($this->computePath($key));
```
NiR-: We don't aim to support directory as there's pretty much only Local adapter that has native support for it, thus a "directory" can't exist. What we'll rather support is a mock of it: directory will be only prefixes, so your fix is really valid as we need to delete directory recursively.

In the case of this specific change I'd prefer to not introduce it as it makes an unnecessary extra call for something we don't care.

src/Gaufrette/Adapter/Local.php
```diff
@@ -309,4 +315,41 @@ protected function createDirectory($directory)
...
+    /**
+     * Removes files or directories.
+     *
+     * @see https://github.com/symfony/filesystem/blob/master/Filesystem.php#L159
+     *
+     * @param string|array|\Traversable $files A filename, an array of files, or a \Traversable instance to remove
+     *
+     * @return bool
+     */
+    private function doDelete($files)
```
NiR-: It has been implemented already [here](https://github.com/KnpLabs/Gaufrette/blob/master/tests/Gaufrette/Functional/LocalDirectoryDeletor.php), using RecursiveIterator which is cleaner. Could you rather reuse it please?

NiR-:
> @nicolasmure @NiR- Currently, the fix changes the current behaviour from returning false when trying to delete a non empty directory to deleting the directory recursively.
Do you accept this or do you want to keep the old behaviour ?

Indeed it would be better submit this patch against 1.0 branch.

NiR-:
@Lctrs Any news?

dmkuznetsov:
Hi guys, any news?

Lctrs:
Closing this one in favor of #582 .

PR #528, opened 2017-09-18
Drop HHVM support + sync docker conf with Travis

Lctrs:
Dropping HHVM support because : 
- Tests never worked for it
- MongoDB dropped support for it in their official drivers
- More and more incompatibilites with PHP
- Lack of latest HHVM versions in Travis
- etc.


Lctrs:
@NiR- It's green :)

docker/Dockerfile-php56
```diff
@@ -1,23 +1,55 @@
...
+RUN set -xe \
+    && apk add --no-cache --virtual .build-deps \
+        $PHPIZE_DEPS \
+        libssh2-dev \
+        zlib-dev \
+    && docker-php-ext-install \
+        zip \
+    && pecl install \
+        apcu-${APCU_VERSION} \
+        mongodb-${MONGODB_VERSION} \
```
NiR-: Why do you specify versions here? We should always install the most up-to-date version.

docker/Dockerfile-php56
```diff
@@ -1,23 +1,55 @@
...
+        zlib
 
-COPY composer.json ${SRC_DIR}/
-RUN sudo chown -R travis ${SRC_DIR} && \
-    composer global require hirak/prestissimo && \
-    composer update --prefer-lowest --prefer-stable && \
-    composer require microsoft/windowsazure:~0.4.3 && \
-    rm -rf ~/.composer/cache/*
+# PHP extensions
+ENV APCU_VERSION="4.0.11" \
```
NiR-: Do you plan to clean/fix the tests for the Apcu adapter? So far we didn't want to dedicate time to maintain deprecated adapters. It's not the best but we don't merge BC changes, so it's not really a problem.

NiR-: My bad, we don't even have an Apcu adapter. There's the Apc one, but it's deprecated so we deprecated the adapter too. It'll be removed in the next major.

docker/Dockerfile-php71
```diff
@@ -1,29 +1,55 @@
...
-    rm -rf ~/.composer/cache/*
-
-VOLUME ${SRC_DIR}/vendor
-
-COPY . ${SRC_DIR}/
-CMD "${SRC_DIR}/bin/tests"
+        zlib
+
+# PHP extensions
+ENV APCU_VERSION="5.1.8" \
```
NiR-: Same here: apcu is not required.

docker/Dockerfile-php70
```diff
@@ -1,28 +1,55 @@
...
-    rm -rf ~/.composer/cache/*
-
-VOLUME ${SRC_DIR}/vendor
-
-COPY . ${SRC_DIR}/
-CMD "${SRC_DIR}/bin/tests"
+        zlib
+
+# PHP extensions
+ENV APCU_VERSION="5.1.8" \
```
NiR-: And here too: apcu is not required.

docker/docker-php-entrypoint.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+set -e
+
+composer update --prefer-dist --no-progress --no-suggest --no-interaction
```
NiR-: Usually I don't like this sort of entrypoint because it means every time I'll fire up a new container it'll run composer. I guess you are used to start a new shell in your containers, but in my case usually I do `docker-compose run --rm php70 phpspec run`. In the other hand the set of depencencies won't change often, so I'd prefer to re-add this in the Dockerfile.

docker-compose.yml
```diff
@@ -1,72 +1,52 @@
...
     build:
-      context: .
-      dockerfile: docker/Dockerfile-php56
+      context: docker
+      dockerfile: Dockerfile-php56
     volumes:
-      - './spec/:/usr/src/gaufrette/spec/'
-      - './src/:/usr/src/gaufrette/src/'
-      - './tests/:/usr/src/gaufrette/tests/'
+      - './:/usr/src/gaufrette:rw'
```
NiR-: If you do this we won't be able to run multiple php versions at the same time. Currently only `spec`, `src` and `tests` are bind-mounded to not mount `vendor`.

docker/Dockerfile-ftp
```diff
@@ -1,17 +1,3 @@
-FROM ubuntu:precise
+FROM stilliard/pure-ftpd
```
NiR-: That's great :+1: In the end, this container isn't really stable.

docker/Dockerfile-php56
Lctrs: Just in case they release a new version that is broken/incompatible with the current PHP version.

docker-compose.yml
Lctrs: That's why the entrypoint script is here.

It will update the vendor when running the tests so we always have dependencies that are compatible with the runned PHP version.

With this, running the tests is just a one command line : 
```sh
bin/tests-all
```

NiR-: Yeah but it means that we can't run different php versions in parallel during development and we've to reinstall dependencies again and again whereas they don't change. Tests should be fast to execute and no extra noise should be added.

docker/Dockerfile-php56
NiR-: IMHO, that's a best practice for projects but not for library. It's the same problem as we had some years ago with `composer.lock` and php libs. I think we shouldn't force extension versions unless we've good reason to do so, like `ssh2`, which is broken.

NiR-:
@Lctrs Could you fix my last comments please? It would be really beneficial to not mount vendor directory as it would allow to run multiple versions of php in parallel.

Lctrs:
@NiR- Done.

NiR-:
Awesome! Thank you @Lctrs :)

PR #533, opened 2017-09-21
[RFR] [1.x] OpenCloud adapter rewrite

nicolasmure:
Related to #497. Also applies propositions from #521.

Replaces the [`rackspace/php-opencloud`](https://github.com/rackspace/php-opencloud) SDK by the [`php-opencloud/openstack`](https://github.com/php-opencloud/openstack) SDK.

## TODO

- [x] do not create container automatically

## Prevents merging
- [x] Merge #521 and rebase this branch
- [x] PHP 7.2 support (see https://github.com/php-opencloud/openstack/issues/192)
- [x] complete the `@TODO`
- [x] Copy the metadata from to source to the dest when renaming a file

## Todo when ready to merge
- [x] Squash commits


nicolasmure:
Well, tests are broken because of the Filesystem not handling the new implementation of the adapters (ie exceptions and void return functions).
I'll try to update the Filesystem too.

src/Gaufrette/Adapter/OpenStack.php
```diff
@@ -0,0 +1,351 @@
...
+            return $this->container;
+        }
+
+        try {
+            if ($this->objectStore->containerExists($this->containerName)) {
+                return $this->container = $this->objectStore->getContainer($this->containerName);
+            }
+
+            if (!$this->createContainer) {
+                throw new \RuntimeException(sprintf('Container "%s" does not exist.', $this->containerName));
```
NiR-: Could you replace all `\RuntimeException` with `StorageFailure` exception please?

src/Gaufrette/Adapter/OpenStack.php
```diff
@@ -0,0 +1,351 @@
...
+            }
+
+            try {
+                return $this->container = $this->objectStore->createContainer(['name' => $this->containerName]);
+            } catch (BadResponseError $e) {
+                throw new \RuntimeException(
```
NiR-: And it would be better to pass `$e` as `$previous`.

src/Gaufrette/Adapter/OpenStack.php
```diff
@@ -0,0 +1,351 @@
...
+    public function rename($sourceKey, $targetKey)
+    {
+        if ($this->exists($targetKey)) {
+            throw StorageFailure::unexpectedFailure(
+                'rename',
+                ['sourceKey' => $sourceKey, 'targetKey' => $targetKey]
+            );
+        }
+
+        $this->write($targetKey, $this->read($sourceKey));
```
NiR-: Unfortunately, exceptions thrown by `write` and `delete` methods are specific to those methods. Thus, I think it would be better to wrap both calls in their own try/catch block.

spec/Gaufrette/Adapter/OpenStackSpec.php
```diff
@@ -0,0 +1,429 @@
...
+
+    function it_throws_storage_failure_while_fetching_metadata(Container $container, StorageObject $object)
+    {
+        $container->getObject('test')->willReturn($object);
+        $object->getMetadata()->willThrow($this->getBadResponseError(400));
+
+        $this->shouldThrow('Gaufrette\Exception\StorageFailure')->duringgetMetadata('test');
+    }
+
+    function it_set_metadata(Container $container, StorageObject $object)
```
NiR-: `it_sets` ;)

spec/Gaufrette/Adapter/OpenStackSpec.php
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    function it_throws_storage_failure_while_listing_objects(Container $container)
+    {
+        $container->listObjects()->willThrow($this->getBadResponseError(400));
+
+        $this->shouldThrow('Gaufrette\Exception\StorageFailure')->duringkeys();
+    }
+
+    function it_list_objects_with_prefix(Container $container)
```
NiR-: `it_lists`

spec/Gaufrette/Adapter/OpenStackSpec.php
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    function it_throws_storage_failure_while_checking_if_file_exists(Container $container)
+    {
+        $container->objectExists('test')->willThrow($this->getBadResponseError(400));
+
+        $this->shouldThrow('Gaufrette\Exception\StorageFailure')->duringexists('test');
+    }
+
+    function it_list_objects(Container $container)
```
NiR-: Same here

src/Gaufrette/Adapter/OpenStack.php
```diff
@@ -0,0 +1,351 @@
...
+            }
+
+            throw StorageFailure::unexpectedFailure('delete', ['key' => $key], $e);
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function rename($sourceKey, $targetKey)
```
NiR-: You need to check if source file exists too.

spec/Gaufrette/Adapter/OpenStackSpec.php
```diff
@@ -0,0 +1,429 @@
...
+    {
+        $client = new Client();
+        $api = new Api();
+
+        $generate = function () use ($client, $api) {
+            for ($i = 0; $i < 3; $i++) {
```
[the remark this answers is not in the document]
nicolasmure: I have to yield `SorageObject` objects, not only simple strings :/

src/Gaufrette/Adapter/OpenStack.php
nicolasmure: done ;)

.travis.yml
```diff
@@ -26,9 +26,9 @@ env:
 
 matrix:
   include:
-    - php: '7.1'
-    - php: '7.2'
-    - php: '7.3'
+    - php: '7.1.29'
```
NiR-: Why do we require specific patch versions? IMHO we should always build using last patch version.

.env.dist
```diff
@@ -21,6 +21,11 @@ AZURE_ACCOUNT=
...
 
-RACKSPACE_USER=
-RACKSPACE_APIKEY=
-RACKSPACE_CONTAINER=
+RACKSPACE_USERNAME=
+RACKSPACE_PASSWORD=
+RACKSPACE_TENANT_ID=
+RACKSPACE_REGION=
+
+IBMCLOUD_USERID=
```
NiR-: Are these credentials set on Travis?

src/Gaufrette/Adapter/OpenStack.php
```diff
@@ -0,0 +1,377 @@
...
+     * @var Container
+     */
+    private $container;
+
+    /**
+     * @param Service $objectStore
+     * @param string  $containerName   The name of the container
+     * @param bool    $createContainer Whether to create the container if it does not exist
+     */
+    public function __construct(Service $objectStore, $containerName, $createContainer = false)
```
NiR-: Could you add proper typehints here please?

src/Gaufrette/Adapter/OpenStack.php
```diff
@@ -0,0 +1,377 @@
...
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function rename($sourceKey, $targetKey)
+    {
+        if (!$this->exists($sourceKey) || $this->exists($targetKey)) {
+            throw StorageFailure::unexpectedFailure(
```
NiR-: Not sure how other adapters do, but I think a `FileNotFound` exception would be more suited. What do you think?

src/Gaufrette/Adapter/OpenStack.php
```diff
@@ -0,0 +1,377 @@
...
+        try {
+            if ($this->objectStore->containerExists($this->containerName)) {
+                return $this->container = $this->objectStore->getContainer($this->containerName);
+            }
+
+            if (!$this->createContainer) {
+                throw new StorageFailure(sprintf('Container "%s" does not exist.', $this->containerName));
+            }
+
+            return $this->container = $this->createContainer();
```
NiR-: I'm not sure about this one too, I guess there're some security-related configuration that might be done on OpenStack storage containers :thinking:

nicolasmure: So better drop the auto creation of containers and let the user do it on its side ?
(ie drop the `createContainer` flag from the adapters constructor)
Seems reasonable indeed. Let me know if you agree and I'll update the roadmap for v1.

.env.dist
nicolasmure: Yes they are

.travis.yml
nicolasmure: Travis had issues with curl on its minor php versions. I tried to manually download a root certificate and to set a php option to point to this certificate but there was no changes.
Secifying the latest patch fixed the issue (see commit https://github.com/KnpLabs/Gaufrette/pull/533/commits/9224c747769dd945e1e5a5eb2dbd67c942383b20 ).

I can try to always use the latest version instead of hardcoding it.

src/Gaufrette/Adapter/OpenStack.php
NiR-: iirc we already discussed that but for the record: yeah, this should be removed from all the adapters.

nicolasmure: Yes :+1: it has been added to the v1 roadmap (#618)

chadyred:
Very nice job guys !

PR #538, opened 2017-12-15
Fixed typo

dawkaa:
Fixed typo in the README and removed version constraint from the
composer description because it no longer holds.


NiR-:
Thank you :)

PR #541, opened 2017-12-15
Quickfix for Adapter/AwsS3, check if count() call is allowed.

andreasschacht:
(cherry picked from commit ad3aad3)
Quickfix for issue #539 

In my case, there was no key `Content` in the `$result`-object. So I check.

PR #544, opened 2018-01-04
Quickfix for Adapter/AwsS3, check if count() call is allowed.

### LINKED ISSUE #539: AwsS3 Adapter isDirectory() fails
https://github.com/KnpLabs/Gaufrette/blob/v0.4.0/src/Gaufrette/Adapter/AwsS3.php#L272
result in an Exception with PHP 7.2:
[object] (RuntimeException(code: 0): count(): Parameter must be an array or an object that implements Countable at /var/www/symfony/vendor/knplabs/gaufrette/src/Gaufrette/Adapter/AwsS3.php:235)

nicolasmure:
Replaces #541 
Fixes #539 

Credits to @andreasschacht


nicolasmure:
Thank you @andreasschacht for your contribution !

PR #549, opened 2018-01-08
Fix a race condition in AzureBlobStorage::keys() when used in multi container mode

NiR-:
That should fix Travis builds :)

This race condition ocurrs when the adapter is used in multi container mode. When keys() is called,
containers are listed and iterated, but if one of them is deleted before its content is listed,
then the adapter throws an exception when iterating over it.


nicolasmure:
Thanks !!

PR #551, opened 2018-01-08
Drop SafeLocal adapter

NiR-:
Even if this adapter provides quite useful feature, it does not fit well in the ongoing design overhaul.

It would be better to have a dedicated "Path transformer" concept so base64 transformation could be generalized to other adapters. For instance, it would also make possible to plug (custom) shard logic to any adapter.

Eventually, we have to deprecate this adapter in the 0.x branch, but only when we have something to supersede it.


nicolasmure:
Can you also indicate in the [1.0 project](https://github.com/KnpLabs/Gaufrette/projects) that this adapter has to be superseded please ?

NiR-:
Oh right, I did not thought about project notes. I don't create any issue for now because I'm not even sure where this feature should live in (either here or in the extras package).

PR #547, opened 2018-01-05
[1.x] Drop support of PHP 5.6

NiR-:
Related to #497.

PR #552, opened 2018-01-12
[1.x] bump to php7.0

nicolasmure:
related to #547


nicolasmure:
I had to enable the `ftp` extension as it is [not enabled by default on windows for php 7](https://stackoverflow.com/questions/36694483/php-7-0-5-use-of-undefined-constant-ftp-binary-assumed-ftp-binary). The other diff lines are due to extension names sort.

PR #566, opened 2018-04-18
Add PutObjectAcl in the required permission

tristanbes:
Without this permission, the application can't create a media in the bucket.


nicolasmure:
Thank you @tristanbes !

PR #570, opened 2018-04-25
docs fix

jakob-stoeck:
`AwsS3PresignedUrlResolver` was once incorrectly written as `AwsS3PresignedResolver`, it also needs 4 parameters. Switched example to `AwsS3PublicUrlResolver`.


nicolasmure:
Good catch @jakob-stoeck !

Issue #575, opened 2018-08-30
FilesystemMap::set should use Filesystem interface instead of concrete implementation


clement-michelet:
The filesystem map depends on concrete implementation instead of interface.
Because of that, It is not possible to use another filesystem implementation like Resolvable filesystem.

PR #576, opened 2018-08-30
FilesystemMap::set() should expects FilesystemInterface instead of Filesystem

clement-michelet:
Fix: #575


nicolasmure:
Thank you @clement-michelet :+1: !

PR #580, opened 2018-09-07
[RFR] fix docker setup for dev env

nicolasmure:
I was not able to build the images from the current docker setup :
```
/tmp/pear/temp/mongodb/src/libmongoc/src/mongoc/mongoc-stream-tls-openssl-bio.c:71:1: error: static declaration of 'BIO_set_init' follows non-static declaration
 BIO_set_init (BIO *b, int init)
 ^~~~~~~~~~~~
In file included from /tmp/pear/temp/mongodb/src/libmongoc/src/mongoc/mongoc-stream-tls-openssl-bio-private.h:27:0,
                 from /tmp/pear/temp/mongodb/src/libmongoc/src/mongoc/mongoc-stream-tls-openssl-bio.c:31:
/usr/include/openssl/bio.h:610:7: note: previous declaration of 'BIO_set_init' was here
 void  BIO_set_init(BIO *a, int init);
       ^~~~~~~~~~~~
make: *** [Makefile:503: src/libmongoc/src/mongoc/mongoc-stream-tls-openssl-bio.lo] Error 1
ERROR: `make' failed
ERROR: Service 'php71' failed to build: The command '/bin/sh -c set -xe     && apk add --no-cache --virtual .build-deps         $PHPIZE_DEPS         libssh2-dev         zlib-dev     && docker-php-ext-install         zip     && pecl install         mongodb-${MONGODB_VERSION}         ssh2-${SSH2_VERSION}     && docker-php-ext-enable --ini-name 05-opcache.ini opcache     && docker-php-ext-enable         mongodb         ssh2     && apk del .build-deps' returned a non-zero code: 1
```
so I rewrote the docker setup for dev env.

The main change with the previous setup is that the dependencies are not installed in the image any more, but directly on the host machine as the whole project is mounted on `/app` in the container (see the `docker-compose.yml` file). Doing so, it's easier to look at the source code of vendors with your text editor.

The drawback is that it's not as easy as before to switch from a php version to another (and so to launch the tests on various php versions). There's an instruction in the `README` which explains how to switch between php versions. But afterall, I'm not sure that it's gonna be a widely encountered use case, and the various php versions are covered by the CI test matrix anyway.

Let me know if you agree.

## TODO

- [x] rebase the fixup commits


nicolasmure:
@NiR- I updated this PR for the php7.1 bump. However, the php 7.3 image fails to build as the [ssh2 extension is not available for php 7.3 yet](https://serverpilot.io/docs/how-to-install-the-php-ssh2-extension).

docker/php/Dockerfile
```diff
@@ -0,0 +1,107 @@
...
+RUN apk add --no-cache \
+        autoconf \
+        g++\
+        git \
+        make \
+        zlib-dev \
+    && pecl install \
+        mongodb \
+    && apk add --no-cache \
+        libssh2-dev \
```
NiR-: Why don't you merge this `apk add` with the previous?

docker/php/Dockerfile
```diff
@@ -0,0 +1,107 @@
...
+    && pecl install \
+        ssh2-1.1.2
```
NiR-: Same here

docker/php/Dockerfile
```diff
@@ -0,0 +1,107 @@
...
+
+RUN docker-php-ext-install \
+        zip \
+    && docker-php-ext-enable \
+        mongodb \
+        ssh2 \
+        zip
```
NiR-: iirc no need to enable an extension installed with `docker-php-ext-install` :wink:

PR #582, opened 2018-09-13
Fix directory deletion

athos7933:
Next to this PR #526


athos7933:
@NiR- @nicolasmure  i rebase my PR

src/Gaufrette/Adapter/Local.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Gaufrette\Adapter;
 
+use Gaufrette\Functional\LocalDirectoryDeletor;
```
nicolasmure: This class is defined under the `Gaufrette\Functional` namespace, which is used for the [tests](https://github.com/KnpLabs/Gaufrette/blob/39b4e3d2ac7c50574d0f9b501342576a8236bc54/tests/Gaufrette/Functional/LocalDirectoryDeletor.php). Maybe you could move this function into the `Local` adapter itself and then use it in this class (and also update the tests to use the moved function).

athos7933:
@NiR- @nicolasmure Any news ?

src/Gaufrette/Adapter/Local.php
```diff
@@ -309,4 +315,29 @@ protected function createDirectory($directory)
             throw new \RuntimeException(sprintf('The directory \'%s\' could not be created.', $directory));
         }
     }
+
+    public static function deleteDirectory($directory)
```
NiR-: This method should be private and tests should use the filesystem to delete their base directory.

athos7933: Sorry but i can't remove static. In tests i can't use the method `$this->filesystem->delete($this->directory)` because the filesystem checks file with this method `assertHasFile`and not working with directory.

athos7933:
@NiR- @nicolasmure Any news ?

nicolasmure:
I cherry picked the commit and applied the changes in #610 .

Thank you @athos7933 !

PR #587, opened 2018-10-27
add and fix php 7.3

andreybolonin:
Travis build done.

PR #591, opened 2018-12-08
fix php7.3 warning about continue in switch

broncha:
In Path.php line 31:

  [Symfony\Component\Debug\Exception\ContextErrorException]
  Warning: "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"?

### DIFF
--- a/src/Gaufrette/Util/Path.php
+++ b/src/Gaufrette/Util/Path.php
@@ -27,13 +27,13 @@ class Path
         foreach ($parts as $part) {
             switch ($part) {
                 case '.':
-                    continue;
+                    break;
                 case '..':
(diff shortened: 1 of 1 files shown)


nicolasmure:
Hello @broncha :)

Thank you for your contribution :) There is already a PR covering this fix : https://github.com/KnpLabs/Gaufrette/pull/587 . I'm going to close yours and use the other one instead, now that php 7.3 has been released.

However, we appreciate your effort here ;)

Bye :)

broncha:
@nicolasmure No problem. Btw should it be break or continue 2?

nicolasmure:
As we want to continue the `foreach` loop, the instruction should be `continue 2;` ;)

PR #593, opened 2018-12-10
add and fix php 7.3

### LINKED ISSUE #592: ContextErrorException when using PHP 7.3.0
Since switching to PHP 7.3.0, I'm getting the following error when using Gaufrette 0.6.0 and 0.7.0:

```PHP Fatal error:  Uncaught Symfony\Component\Debug\Exception\ContextErrorException: Warning: "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in vendor/knplabs/gaufrette/src/Gaufrette/Util/Path.php:30```

Looks like changing https://github.com/KnpLabs/Gaufrette/blob/master/src/Gaufrette/Util/Path.php#L30 from `continue` to `break` might be enough to fix this.

### PR DESCRIPTION
**Pull request by nicolasmure**
I cherry-picked the commit from #587 cc/ @andreybolonin 
Fixes #592 .

### CODE REVIEW

File: .travis.yml [hunk 1]
```diff
@@ -30,6 +30,7 @@ matrix:
     - php: '7.0'
```
NiR-: I think we should also drop support for PHP 5.6 and 7.0, as 7.0 has been EOL'ed and 5.6 will be at the end of the year.

deguif:
Is a new release planned with these fixes?

nicolasmure:
Yes it is (see https://github.com/KnpLabs/Gaufrette/pull/597), we'll try to also make a new release of the bundle.

deguif:
Nice, thanks you very much @nicolasmure

nicolasmure:
@deguif [v0.8.0](https://github.com/KnpLabs/Gaufrette/releases/tag/v0.8.0) is now out ;)

PR #602, opened 2019-01-26
Fixing GridFS metadata bug where no metadata array returned

bsperduto:
Fixing a gridfs adapter bug where the object exists but it does not contain a metadata array.


nicolasmure:
Hello @bsperduto :)
Thank you for your contribution ! Do you mind to write a spec covering this case ?

bsperduto:
> Hello @bsperduto :)
> Thank you for your contribution ! Do you mind to write a spec covering this case ?

Added a phpunit test to cover it. Gaufrette always creates the metadata object but when an outside library creates the gridfs file it doesn't always exist. This should cover that situation.

nicolasmure:
Replaced by #609 (cherry-picked)

PR #607, opened 2019-04-30
Fix for the isDir function in the FTP Adapter

Romain:
The isDir function seems to be broken in the FTP Adapter.
When I try to retrieve a document through a stream, I get this error: `ftp_chdir(): /documents/your-file-name.jpg: No such file or directory`

Here is the complete trace:

```
in vendor/knplabs/gaufrette/src/Gaufrette/StreamWrapper.php->stat (line 210) 
in vendor/knplabs/gaufrette/src/Gaufrette/Stream/InMemoryBuffer.php->isDirectory (line 159) 
in vendor/knplabs/gaufrette/src/Gaufrette/Filesystem.php->isDirectory (line 333)
in vendor/knplabs/gaufrette/src/Gaufrette/Adapter/Ftp.php->isDir (line 222)
in vendor/knplabs/gaufrette/src/Gaufrette/Adapter/Ftp.phpftp_chdir (line 369)
in vendor/knplabs/gaufrette/src/Gaufrette/Adapter/Ftp.php (line 369)
```

Changing the way we perform this test by using the PHP function `is_dir` instead of the `ftp_chdir` one fixes this error.


src/Gaufrette/Adapter/Ftp.php
```diff
@@ -366,7 +366,12 @@ private function isDir($directory)
             return true;
         }
 
-        if (!@ftp_chdir($this->getConnection(), $directory)) {
```
nicolasmure: Thank you for your contribution :)

From what I read on the php doc [1], the usage of `is_dir` with an URI having the `ftp`/`sftp` protocol is only allowed in passive mode. By default, the passive mode is disabled (c.f. the `__contructor` of this adapter).

From your PR description, it looks like you have an exception. What about to still use the `ftp_chdir` function, but catch the exception to fallback to `is_dir` ?

Also, can you create a dedicated method to compute the connection string ?

[1] https://php.net/manual/en/wrappers.ftp.php

Romain: Thanks for these tips!
I created a private method for the connection URL.
I also kept the `chdir` method as the main approach and fallback on `is_dir` with passive mode set to `true` if we catch and exception with `chdir`.
Let me know if this looks better to you.

Romain (commented):
I created a private method for the connection URL.
I also kept the chdir method as the main approach and fallback on is_dir with passive mode set to true if we catch and exception with chdir.

src/Gaufrette/Adapter/Ftp.php
```diff
@@ -366,7 +377,22 @@ private function isDir($directory)
...
+        $chDirResult = false;
+        try {
+            $chDirResult = ftp_chdir($this->getConnection(), $directory);
```
nicolasmure: can you also move the change to the original directory here ?
```php
// change directory again to return in the base directory
ftp_chdir($this->getConnection(), $this->directory);
```
as it'd keep the `ftp_chdir` logic in this `try` block :

```php
try {
    $isDir = ftp_chdir($this->getConnection(), $directory);
    // change directory again to return in the base directory
    ftp_chdir($this->getConnection(), $this->directory);

    return $isDir;
}
```

nicolasmure (commented):
Thanks for the update ;)

src/Gaufrette/Adapter/Ftp.php
```diff
@@ -366,7 +377,22 @@ private function isDir($directory)
...
+        }
+        catch(Exception $e) {
+            $this->passive = true;
```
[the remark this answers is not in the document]
Romain: If the `ftp_chdir` function generates an exception, the idea is to fallback on the `is_dir` function. Therefore, to have it working, I'm forcing the passive mode to `TRUE`.
I think that I should revert back this value to its original state though. Pushing a fix for this.

Romain:
@nicolasmure I pushed a fix with your comments applied in it.

src/Gaufrette/Adapter/Ftp.php
```diff
@@ -366,12 +378,36 @@ private function isDir($directory)
...
+        $chDirResult = false;
+        try {
+            $chDirResult = ftp_chdir($this->getConnection(), $this->directory);
+
+            // change directory again to return in the base directory
+            ftp_chdir($this->getConnection(), $this->directory);
+        } catch(\Exception $e) {
+
+            // Cache the actual value of the passive mode
+            $actualPassiveMode  = $this->passive;
```
nicolasmure: Actually, I don't think you have to change this flag. It should be RO, and it is only used to specify whether the backend supports passive mode. See its usage [1] for more details. Changing the value of the flag do not brings any change in the behavior of the PHP's FTP functions. The flag is only used as an indicator in this adapter.

So in the end, the call to `is_dir` should be made only when the passive mode is enabled. When it is not enabled, you can throw the `$e` exception as we can't use `is_dir` :

> If the server does not support passive mode ftp, the connection will fail. 

*[2]*

Also, it could be good to add a comment with the link to [2] above the usage of `is_dir` and the check of the passive mode to ease understanding.

[1] https://github.com/KnpLabs/Gaufrette/blob/350f0f68e72c9f39845413c9b775ac49ffa18e8b/src/Gaufrette/Adapter/Ftp.php#L548
[2] https://php.net/manual/en/wrappers.ftp.php

nicolasmure (commented):
Can you also squash your commits please ? :)

Romain:
@nicolasmure Got it. Pushed a change accordingly to your comment.

src/Gaufrette/Adapter/Ftp.php
```diff
@@ -366,12 +378,31 @@ private function isDir($directory)
...
+            }
+
+            if (!@is_dir($url . $directory)) {
+                return false;
+            }
         }
 
-        // change directory again to return in the base directory
-        ftp_chdir($this->getConnection(), $this->directory);
+        if (!$chDirResult) {
```
nicolasmure: In the end this condition can be avoided (as the final `return true`).
You can directly `return $chDirResult;` after the call to `ftp_chdir` to get back to the original dir ;)

src/Gaufrette/Adapter/Ftp.php
```diff
@@ -366,12 +378,31 @@ private function isDir($directory)
             return true;
         }
 
-        if (!@ftp_chdir($this->getConnection(), $directory)) {
-            return false;
+        $chDirResult = false;
```
nicolasmure: This line can be removed ;)
The scope of this var should be in the `try` block only.

src/Gaufrette/Adapter/Ftp.php
```diff
@@ -366,12 +378,31 @@ private function isDir($directory)
...
+        try {
+            $chDirResult = ftp_chdir($this->getConnection(), $this->directory);
+
+            // change directory again to return in the base directory
+            ftp_chdir($this->getConnection(), $this->directory);
+        } catch(\Exception $e) {
+
```
nicolasmure: please remove this empty line ;)

src/Gaufrette/Adapter/Ftp.php
```diff
@@ -366,12 +378,31 @@ private function isDir($directory)
...
+            // Build the FTP URL that will be used to check if the path is a directory or not
+            $url = $this->createConnectionUrl();
```
nicolasmure: The `$url` var should be declared after the check on the passive mode, otherwise it may be declared without being used.

src/Gaufrette/Adapter/Ftp.php
```diff
@@ -366,12 +378,31 @@ private function isDir($directory)
...
+
+            // is_dir is only available in passive mode.
+            // See https://php.net/manual/en/wrappers.ftp.php for more details.
+            if(!$this->passive) {
```
nicolasmure: please keep a space after the `if` statement and the opening parenthesis (PSR)

src/Gaufrette/Adapter/Ftp.php
```diff
@@ -366,12 +378,31 @@ private function isDir($directory)
...
 
-        if (!@ftp_chdir($this->getConnection(), $directory)) {
-            return false;
+        $chDirResult = false;
+        try {
+            $chDirResult = ftp_chdir($this->getConnection(), $this->directory);
+
+            // change directory again to return in the base directory
+            ftp_chdir($this->getConnection(), $this->directory);
+        } catch(\Exception $e) {
```
nicolasmure: same here : can you add a space between the `catch` statement and the opening parenthesis please ?

src/Gaufrette/Adapter/Ftp.php
```diff
@@ -366,12 +378,31 @@ private function isDir($directory)
...
+            // Build the FTP URL that will be used to check if the path is a directory or not
+            $url = $this->createConnectionUrl();
+
+            // is_dir is only available in passive mode.
+            // See https://php.net/manual/en/wrappers.ftp.php for more details.
+            if(!$this->passive) {
+                throw new \BadFunctionCallException('is_dir can only be used in passive mode.');
+            }
+
+            if (!@is_dir($url . $directory)) {
```
nicolasmure: can't we directly `return @is_dir(/* ... */);` ?

src/Gaufrette/Adapter/Ftp.php
```diff
@@ -366,12 +378,31 @@ private function isDir($directory)
...
+            ftp_chdir($this->getConnection(), $this->directory);
+        } catch(\Exception $e) {
+
+            // Build the FTP URL that will be used to check if the path is a directory or not
+            $url = $this->createConnectionUrl();
+
+            // is_dir is only available in passive mode.
+            // See https://php.net/manual/en/wrappers.ftp.php for more details.
+            if(!$this->passive) {
+                throw new \BadFunctionCallException('is_dir can only be used in passive mode.');
```
nicolasmure: From the docblocks of the other methods using this `isDir` function, they can only throw `\RuntimeException`.
What about changing this line to a `RuntimeException` with a more explicit error message :
```php
throw new \RuntimeException(
    \sprintf('Not able to determine whether "%s" is a directory or not. Please try again using a passive FTP connection if your backend supports it, by setting the "passive" option of this adapter to true.', $directory),
    $e->getCode(),
    $e
);
```

nicolasmure (commented):
Thanks for the changes :)

There are also some failing tests, do you mind to fix them ?

phpunit :
```
There were 2 failures:
1) Gaufrette\Functional\Adapter\FtpTest::shouldCheckIfFileExists
Failed asserting that true is false.
/home/travis/build/KnpLabs/Gaufrette/tests/Gaufrette/Functional/Adapter/FunctionalTestCase.php:91

2) Gaufrette\Functional\Adapter\FtpTest::shouldFailWhenTryMtimeForKeyWhichDoesNotExist
Failed asserting that 1561017417 is false.
/home/travis/build/KnpLabs/Gaufrette/tests/Gaufrette/Functional/Adapter/FunctionalTestCase.php:119
```
and also the phpspec for this adapter.

You can run the specs with
```bash
$ docker/run-task php7.2 vendor/bin/phpspec run spec/Gaufrette/Adapter/Ftp.php
```
and the integration tests with
```bash
$ docker/run-task php7.2 vendor/bin/phpunit tests/Gaufrette/Functional/Adapter/FtpTest.php
```

src/Gaufrette/Adapter/Ftp.php
Romain: Gosh, true. I didn't see this anymore.

src/Gaufrette/Adapter/Ftp.php
Romain: Correct.

src/Gaufrette/Adapter/Ftp.php
Romain: Yes, that's a nice simplification.

Romain:
Your comments have been integrated. These are nice improvements. ;)

src/Gaufrette/Adapter/Ftp.php
```diff
@@ -378,30 +378,27 @@ private function isDir($directory)
             return true;
         }
 
-        $chDirResult = false;
         try {
+            $chDirResult = false;
```
nicolasmure: This line can be removed as the boolean value is set by the call to `ftp_chdir` :

> Returns TRUE on success or FALSE on failure.

https://www.php.net/manual/en/function.ftp-chdir.php

src/Gaufrette/Adapter/Ftp.php
```diff
@@ -378,30 +378,27 @@ private function isDir($directory)
...
-        }
 
-        if (!$chDirResult) {
-            return false;
+            // Build the FTP URL that will be used to check if the path is a directory or not
+            $url = $this->createConnectionUrl();
+            return @is_dir($url . $directory);
         }
 
         return true;
```
nicolasmure: This line can be removed as this statement won't be reached anymore ;)

nicolasmure (commented):
Can you also update the tests accordingly please ?

src/Gaufrette/Adapter/Ftp.php
Romain: Indeed.

Romain:
@nicolasmure I'm not sure how to update the tests since they're not checking the way we connect  to the server.

src/Gaufrette/Adapter/Ftp.php
```diff
@@ -355,6 +355,18 @@ protected function createDirectory($directory)
         }
     }
 
+    /**
+     * @return string
+     */
+    private function createConnectionUrl()
+    {
+        $url = $this->ssl ? 'sftp://' : 'ftp://';
```
stof, line 363: `sftp` is not the same than `ftps`

PR #609, opened 2019-06-04
[GridFS] Fix : return empty array when object has no metadata

nicolasmure:
Replaces #602 (and resolves conflicts).

Credits @bsperduto


nicolasmure:
Thank you @bsperduto !

PR #610, opened 2019-06-04
[Local Adapter] directory deletion

nicolasmure:
Completes #582 (cherry picked)

As @NiR- pointed out, Gaufrette (and so the Filesystem) do not aim to support directories.
We only support objects identified by keys. However, the directory deletion seems to be a requested feature by the community. This PR completes the requested changes of #582.

As the directory deletion is very specific to the Local Adapter, you have to explicitely retrieve the adapter in order to use it :

```php
$filesystem->getAdapter()->delete($key);
```


nicolasmure:
Thank you @athos7933 !

PR #617, opened 2019-06-18
[RFR] Add PHP Coding Standard check step on CI.

nicolasmure:
Only run the `php-cs-compare` make target on travis ASAP to fail the build when the CS are violated.

Also, expose a `php-cs-fix` make target to fix the violations.

### DIFF
--- a/spec/Gaufrette/Adapter/AclAwareAmazonS3Spec.php
+++ b/spec/Gaufrette/Adapter/AclAwareAmazonS3Spec.php
@@ -48,7 +48,7 @@ class AclAwareAmazonS3Spec extends ObjectBehavior
         $service
             ->set_object_acl('bucketName', 'filename2', \AmazonS3::ACL_PRIVATE)
             ->shouldBeCalled()
-            ->willReturn(new \CFResponse(array(), '', 200))
+            ->willReturn(new \CFResponse([], '', 200))
         ;
         $adapter
             ->rename('filename', 'filename2')
@@ -72,7 +72,7 @@ class AclAwareAmazonS3Spec extends ObjectBehavior
         $service
             ->set_object_acl('bucketName', 'filename2', \AmazonS3::ACL_PRIVATE)
             ->shouldBeCalled()
-            ->willReturn(new \CFResponse(array(), '', 500));
+            ->willReturn(new \CFResponse([], '', 500));
         $adapter
             ->rename('filename', 'filename2')
             ->shouldBeCalled()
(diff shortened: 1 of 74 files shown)


spec/Gaufrette/Adapter/AclAwareAmazonS3Spec.php
```diff
@@ -12,25 +12,25 @@ class AclAwareAmazonS3Spec extends ObjectBehavior
      * @param \Gaufrette\Adapter $adapter
      * @param \AmazonS3 $service
      */
-    function let(Adapter $adapter, AmazonS3 $service)
+    public function let(Adapter $adapter, AmazonS3 $service)
```
NiR-: Actually this fix is wrong IMHO: spec functions don't need `public` visibility keyword

spec/Gaufrette/StreamWrapperSpec.php
```diff
@@ -200,21 +200,21 @@ function it_does_not_get_stat_when_is_not_open()
...
+        $stat = [
+            'dev' => 1,
+            'ino' => 12,
+            'mode' => 0777,
             'nlink' => 0,
-            'uid'   => 123,
-            'gid'   => 1,
-            'rdev'  => 0,
-            'size'  => 666,
+            'uid' => 123,
```
NiR-: IMHO would be better to keep things align here

src/Gaufrette/Adapter/AclAwareAmazonS3.php
```diff
@@ -18,14 +18,15 @@
  *
  * @deprecated The AclAwareAmazonS3 adapter is deprecated since version 0.4 and will be removed in 1.0. Use the AwsS3 adapter instead.
  */
-class AclAwareAmazonS3 implements Adapter,
-                                  MetadataSupporter
+class AclAwareAmazonS3 implements
+    Adapter,
+    MetadataSupporter
```
NiR-: Maybe just put everything on a single line? :thinking:

NiR-:
What about using [PedroTroller/PhpCSFixer-Custom-Fixers](https://github.com/PedroTroller/PhpCSFixer-Custom-Fixers#pedrotrollerphpspec) and keep a single php_cs config file?

spec/Gaufrette/StreamWrapperSpec.php
nicolasmure: Actually I'm not fan of aligning things, as it will make changes on other lines as soon as a big line is introduced ; and thus, making a new commit for these lines, whereas they were not significantly changed.
To me, it's the same issue than the trailing comma at the end of array items. Before if was supported, adding a new item to a list was producing a git diff of two lines : the actual line and the above one to add a comma. Only the actual line had a sense for the git diff.

That's why I prefer not to keep things aligned. Only alphabetic sort should improve readability IMHO.

PR #633, opened 2020-01-03
Add test for rename url not safe source

alexgt9:


tests/Gaufrette/Functional/Adapter/AwsS3Test.php
```diff
@@ -188,4 +188,16 @@ public function shouldUploadWithGivenContentType()
...
+
+    /**
+     * @test
+     */
+    public function shouldRenameAnObject()
+    {
+        $this->filesystem->write('foo', '');
+        $this->filesystem->rename('foo', 'foo%_encode');
+
+        $this->assertTrue($this->filesystem->has('foo'));
```
nm2107: shouldn't the original name not be present ?

alexgt9: You are right 👌 
Anyway there is no environment where this tests for S3 are executed?

Nek- (changes requested):
Hello, thank you very much for trying to improve Gaufrette.

I didn't test your fix but I see that the test passed without the change you did in the code.

PR #636, opened 2020-03-15
Remove reset call

chadyred:


chadyred:
As we see an existing test fail, I let here as it is in order to check why this method has any trouble.

PR #650, opened 2021-06-30
[Flystem] Add FlysystemV2 adapter for version 2 support of flysystem.

jenkoian:
Done as a new adapter to allows users to use the existing adapter for
flysystem version 1.

--

Let me know if are ok with the `V2` adapter or if you think we should just overwrite the existing one. 

Haven't touched any phpstan or psalm config, will examine the output of the github actions but may need some pointers here.


jum4:
Nice job @jenkoian, I'm interest in this adapter, maybe I can help,

`league/flysystem:^1.0` seems to be fixed also in `appveyor.yml` and the `Makefile` (used by PHPStan analysis).
Because you kept the flysystem version 1 for compatibilty (which is a good choice in my opinion), you cannot depend on both versions.
Maybe the old Adapter may be excluded from PHPStan analysis ? So the requirement could be replaced by `league/flysystem:^2.0`.

The PHP tests seems to be failed because of trying to install olders versions of PHP (7.1, 7.2, 7.3) on ubuntu-latest, maybe an older version of ubuntu should be required for theses tests.

jum4:
@nicolasmure

jenkoian:
Version 3 of Flysystem out  now too https://github.com/thephpleague/flysystem/releases/tag/3.0.0

dmitryuk:
Finally done in  https://github.com/KnpLabs/Gaufrette/pull/688

PR #656, opened 2022-08-19
:art: Review makefile

Nek-:
Don't get me wrong: it makes things clear, not things working with dockers. This will be the next step :) .

But this fixes the fact that `make` alone is copying the .env and output basically nothing useful. It also clean and fix config files. Optimize docker build. And... prepare the future.

image


Nyholm (approved):
Thank you

PR #658, opened 2022-09-16
⬆️ Starts dependencies & test upgrades and fixes for PHP 8!

Nek-:
There is a lot of work to make Gaufrette work 100% with PHP 8+, it's ridiculous to make it happen in only 1 PR. So here is one part of it.

Please notice that this really fixes async AWS integration, which is lovely if somebody needs it.

Finally, please notice that this also drops compatibility for PHP<7.3.

Replace #657 for convenience (I'm the only one working on all of this, to have separated PRs it's better to work on the repository directly)


spec/Gaufrette/Adapter/FtpSpec.php
```diff
@@ -1,229 +0,0 @@
...
-
-namespace spec\Gaufrette\Adapter;
-
-//hack - mock php built-in functions
-require_once 'functions.php';
-
-use Gaufrette\Filesystem;
-use PhpSpec\ObjectBehavior;
-
-class FtpSpec extends ObjectBehavior
```
Nek-, line 11: Sadly, it's not possible to make a spec for PHP8+ because of the connection object which is impossible to mock.

PR #679, opened 2022-10-10
Fix #649 - Remove useless files from users and servers

### LINKED ISSUE #649: Add a .gitattributes file
image

Remove everything that is related to tools, github, docker, CIs
Keep the tests for people packaging this lib and the phpunit config file. Also keep source files, but nothing more.

williamdes:
Hi !
I did this change in two steps, first I excluded the files from the git archive: `git archive --help`. GitHub uses `git archive` for users to be able to "Download" a branch or git tag.

So I kept the test files and the phpunit config in it because: packing people on Debian/Alpine/Others... will download the source using scripts and do not want to have to clone your git history because they do not care about it and only want the source.
That said this is not what composer users want. They only want source file, so I excluded test files from the bundles that packagist produces.

Final users will only have source files, packaging people will have source+tests and devs can use git clone.
```sh
# Download source from my PR branch & extract
# composer install
~/Gaufrette-gitattrs$ ./vendor/bin/phpunit 
PHPUnit 8.5.30

SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS  63 / 207 ( 30%)
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS........ 126 / 207 ( 60%)
.........SSSSSSSSSSS........................................... 189 / 207 ( 91%)
..................                                              207 / 207 (100%)

Time: 131 ms, Memory: 12.00 MB

OK, but incomplete, skipped, or risky tests!
Tests: 207, Assertions: 170, Skipped: 129.
```
Fixes: #649


composer.json
```diff
@@ -42,5 +42,11 @@
         "branch-alias": {
             "dev-master": "0.10.x-dev"
         }
+    },
+    "archive": {
+        "exclude": [
+            "/tests",
+            "/phpunit.xml.dist"
```
Nek-, line 49: spec folder and phpspec config file should be here to follow the logic entirely (it's part of the test suite). The .env.dist file is also absolutely required, as well as the make file.

Nek- (changes requested):
Thank you for your help on improving Gaufrette. This could be part of the next release (coming soon) but I think you may have excluded too much files if I follow the logic of your PR description.

williamdes:
> Thank you for your help on improving Gaufrette. This could be part of the next release (coming soon) but I think you may have excluded too much files if I follow the logic of your PR description.

Very cool !
I figured out that the Makefile without the docker files would not make sense so I also moved it for them. Should be okay now

Nek-:
> I figured out that the Makefile without the docker files would not make sense so I also moved it for them. Should be okay now

TBH there's a bit of cleanup to do about this... Many config files and build-related stuff are sooo outdated... Maybe gonna work on this.

But your work is nice thank you.

Nek- (approved):

williamdes:
> > I figured out that the Makefile without the docker files would not make sense so I also moved it for them. Should be okay now
> 
> TBH there's a bit of cleanup to do about this... Many config files and build-related stuff are sooo outdated... Maybe gonna work on this.
> 
> But your work is nice thank you.

Thank you, feel free to ping me for review on build stuff

Nek- (approved):

Nek-:
Thank you very much @williamdes !

PR #683, opened 2022-11-01
Fixed azure blob write method

rbochenski:
Fixed: when write function is executed with resource instead of string it fail because stream is already closed when using is_resource;


Nek-:
I'm good with this fix. Thank you very much @rbochenski :).

PR #685, opened 2022-11-15
Ignore some files from vendor

dmitryuk:
Excludes some not necessary files from vendor.
Lines ordered alphabetically


alexpozzi (approved):

dmitryuk:
What next?

PR #688, opened 2022-11-25
Flysysystem V2 and V3

dmitryuk:
Added class FlysystemV2V3  - can be used on both versions.
- Much of time spent to fix analizers.
- Psalm upgraded to `4.30.0` to use new plugin system
- CI flow divided to test and analyse different versions of flysystem

Inspired on #650


dmitryuk:
About Roave we can skip it, because originally Gaufrette changed names of arguments and now its it right.
```
[BC] CHANGED: Parameter 0 of Gaufrette\Adapter\InMemory#isDirectory() changed name from path to key
[BC] CHANGED: Parameter 0 of Gaufrette\Stream\InMemoryBuffer#cast() changed name from castAst to castAs
```

Issue #275, opened 2014-04-09
StreamWrapper full support missing


unusorin:
Hi guys,
I've encountered a problem with Gauffrete StreamWrapper. It seems that `dir_open` (http://www.php.net/manual/en/streamwrapper.dir-opendir.php) method wasn't implemented. This causes calls like `is_dir` and `scandir` to crash.

Case:

```
$map = \Gaufrette\StreamWrapper::getFilesystemMap();
$map->set('vendor', new \Gaufrette\Filesystem(new \Gaufrette\Adapter\Local(__DIR__ . DIRECTORY_SEPARATOR . 'vendor/')));

\Gaufrette\StreamWrapper::register();

var_dump(scandir('gaufrette://vendor/'));
```

Output:

```
PHP Warning:  scandir(gaufrette://vendor/): failed to open dir: "Gaufrette\StreamWrapper::dir_opendir" call failed in test.sandbox.php on line 59
```

Environment:

```
OS: Ubuntu 12.04 x64
PHP Version: PHP 5.5.9-1+sury.org~precise+1 (cli) (built: Feb 13 2014 15:53:53) 
                      Copyright (c) 1997-2014 The PHP Group
                      Zend Engine v2.5.0, Copyright (c) 1998-2014 Zend Technologies
                      with Zend OPcache v7.0.3, Copyright (c) 1999-2014, by Zend Technologies
                      with Xdebug v2.2.3, Copyright (c) 2002-2013, by Derick Rethans
 Gaufrette: dev-master
```

Thanks,
Sorin!

zartdinov:
Warning: stream_set_blocking(): Gaufrette\StreamWrapper::stream_set_option is not implemented!

apsylone:
Same here !

Baachi:
Gaufrette only support file actions. You can't create/scan/unlink a folder with the streamwrapper.

unusorin:
Is this an official answer ?

Baachi:
This is the current state, you can change it with a PR :smile:

rimas-kudelis:
> Gaufrette only support file actions. You can't create/scan/unlink a folder with the streamwrapper.

Not even all file actions are supported. `rename()`, for instance, is not.

PR #689, opened 2022-12-09
Add support for php 8.2

xelaris:
See https://github.com/phpspec/phpspec/pull/1435 for a similar update and for the reason why ci isn't successfully yet, as phpspec does't support 8.2 yet.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       max-parallel: 10
       matrix:
-        php: ['7.4', '8.0', '8.1']
+        php: ['7.4', '8.0', '8.1', '8.2']
 
     steps:
       - name: Set up PHP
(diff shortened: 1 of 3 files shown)


xelaris:
@Nek- thank you for the quick feedback.

We are using the stream integration in our project. Actually, the reason for this PR is, that the StreamWrapper causes the deprecation about the dynamic $context property, which is (wrongly?) considered a direct deprecation by the Symfony PHPUnit bridge.

I found https://github.com/KnpLabs/Gaufrette/issues/275, which mentions some issues. In our simple use-case (file_put_contents with AWS S3 adapter) it seems to work fine though.

Do you have any guidance about what else needs to be done to get Gaufrette compatible with PHP 8.2? I might find the time to have a look at it.

In any case, I guess this PR has to wait for release of phpspec which is compatible with PHP 8.2 for all tests become green eventually. Or do you think this PR should be updated to temporarily require `phpspec/phpspec: "dev-main@dev"`?

toooni:
Is there still a reason to not merge this PR? Except for these deprecation messages, our use case works fine on PHP 8.2.

Nek-:
@toooni there is no maintainer working on this project, sorry.

PedroTroller:
Thank you very much. I took the liberty of validating your PR which remained in draft. We'll put your PR in the next version, v0.12.0

PR #698, opened 2023-04-24
Refacto: Add type on Stream classes

KevinArtus:


src/Gaufrette/Stream.php
```diff
@@ -58,52 +54,41 @@ public function close();
...
     /**
      * Seeks to the specified offset.
-     *
-     * @param int $offset
-     * @param int $whence
-     *
-     * @return bool
      */
-    public function seek($offset, $whence = SEEK_SET);
+    public function seek(int $offset, int $whence = SEEK_SET): bool;
```
PedroTroller, line 62: You can add `SEEK_SET|SEEK_CUR|SEEK_END` as `@param` for `$whence`

src/Gaufrette/Stream.php
```diff
@@ -58,52 +54,41 @@ public function close();
...
     /**
      * Retrieve the underlying resource.
      *
-     * @param int $castAs
-     *
-     * @return mixed using resource or false
+     * @return resource|false using resource or false
      */
-    public function cast($castAs);
+    public function cast(int $castAs);
```
PedroTroller, line 86: `$castAs` can be STREAM_CAST_FOR_SELECT or STREAM_CAST_AS_STREAM

src/Gaufrette/Stream/InMemoryBuffer.php
```diff
@@ -9,28 +9,25 @@
...
-    private $numBytes;
-    private $position;
-    private $synchronized;
+    private Filesystem $filesystem;
+    private string $key;
+    private StreamMode $mode;
+    private string $content;
+    private int $numBytes;
+    private int $position;
+    private bool $synchronized;
```
PedroTroller, line 18: I think only apply this typehint will create some troubles. For example, L191 will create failure instead of `return false` like currently.

PedroTroller (changes requested):

muchafm (approved):

src/Gaufrette/Stream/Local.php
```diff
@@ -12,25 +12,22 @@
  */
 class Local implements Stream
 {
-    private $path;
+    private string $path;
     private $mode;
     private $fileHandle;
```
PedroTroller, line 17: Why not type these properties?

PR #699, opened 2023-04-24
refactor: add typeint to FilesystemMap, StreamWrapper and StreamMode

PedroTroller:


KevinArtus (changes requested):

KevinArtus (changes requested):

KevinArtus (approved):

muchafm (approved):

src/Gaufrette/StreamWrapper.php
```diff
@@ -58,149 +56,126 @@ public static function register($scheme = 'gaufrette')
...
-     * @param int    $flags
-     *
-     * @return mixed
+     * @return array<string, mixed>|false
      *
-     * @todo handle $flags parameter
+     * @TODO handle $flags parameter
      */
-    public function url_stat($path, $flags)
+    public function url_stat(string $path, int $flags): array|bool
```
KevinArtus, line 178: It's seems this method only return array

```
    /**
     * Gathers statistics of the stream.
     *
     * @return array
     */
    public function stat();
```

KevinArtus (changes requested):

PR #700, opened 2023-04-24
Refacto/add type util

KevinArtus:


src/Gaufrette/Util/Size.php
```diff
@@ -41,10 +33,8 @@ public static function fromFile($filename)
      * Returns the size in bytes from the given resource.
      *
      * @param resource $handle
-     *
-     * @return string
      */
-    public static function fromResource($handle)
+    public static function fromResource($handle): int
```
muchafm, line 37: I don't think this one is right!

muchafm (changes requested):

src/Gaufrette/Util/Size.php
KevinArtus: PHP doc specified int
https://www.php.net/manual/en/function.mb-strlen.php

muchafm: I don't see `mb_strlen` used in this function

src/Gaufrette/Util/Checksum.php
```diff
@@ -5,30 +5,22 @@
...
 
     /**
      * Returns the checksum of the specified file.
-     *
-     * @param string $filename
-     *
-     * @return string
      */
-    public static function fromFile($filename)
+    public static function fromFile(string $filename): false|string
```
PedroTroller, line 23: `false` return typehint is only available with php 8.2

src/Gaufrette/Util/Size.php
```diff
@@ -27,12 +23,8 @@ public static function fromContent($content)
 
     /**
      * Returns the size in bytes from the given file.
-     *
-     * @param string $filename
-     *
-     * @return int
      */
-    public static function fromFile($filename)
+    public static function fromFile(string $filename): false|int
```
PedroTroller, line 27: Same here, `false` return typehint is only available with php 8.2

src/Gaufrette/Util/Size.php
```diff
@@ -41,10 +33,8 @@ public static function fromFile($filename)
      * Returns the size in bytes from the given resource.
      *
      * @param resource $handle
-     *
-     * @return string
      */
-    public static function fromResource($handle)
+    public static function fromResource($handle): array|false|int
```
PedroTroller, line 37: Same here

PedroTroller (changes requested):

muchafm (approved):

src/Gaufrette/Util/Size.php
```diff
@@ -41,10 +33,8 @@ public static function fromFile($filename)
      * Returns the size in bytes from the given resource.
      *
      * @param resource $handle
-     *
-     * @return string
      */
-    public static function fromResource($handle)
+    public static function fromResource($handle): array|bool|int
```
PedroTroller, line 37: Why can a size be a bool or a string?

PedroTroller (changes requested):

PedroTroller (approved):

PR #706, opened 2023-10-23
Declare StreamWrapper::context explicitly to avoid dynamic property

nocive:
Under PHP 8.2, the following deprecation warning occurs when using the stream wrapper implementation.

```
Deprecated: Creation of dynamic property Gaufrette\StreamWrapper::$context is deprecated
```


nocive:
:man_facepalming: forgot to check existing open PRs before opening one.

PR #722, opened 2026-07-20
feat!: PHP 8.5 compatibility updates

clementvtrd:
Enhance compatibility with PHP 8.5 by updating dependency requirements and modifying nullable type hints throughout the codebase. This ensures smoother integration and reduces deprecation warnings.

[Roave Backwards Compatibility Check](https://github.com/KnpLabs/Gaufrette/actions/runs/29747840201/job/88370272776?pr=722) is expected to fail as we are dropping supports of PHP 7.


composer.json
```diff
@@ -40,7 +40,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "0.10.x-dev"
+            "dev-fix-php82-deprecated": "0.12.x-dev"
```
garak: This is probably to be reverted

phpspec.yml
```diff
@@ -0,0 +1,4 @@
+# Exclude E_DEPRECATED (8192) from E_ALL (32767): dependencies such as
```
garak, line 1: This file should be added to .gitattributes

clementvtrd: Do you think about something like this?

```yaml
/phpspec.yml export-ignore
```

composer.json
clementvtrd: This should probably entirely remove, no? `master` should be enough to install the dev version

garak: I think it should stay, indeed. Having an alias is always good, because users can leverage the alias and avoid breaking changes when in "dev" mode. This is because while "dev-master" always points to the latest release, the alias is scoped to the next release.

clementvtrd: Roger that, it's reverted 👍

garak (approved):

